PR #21, opened 2019-11-21
[STACK-431] Rack devices

omkartelee01:
As a A10 Octavia user, I must be able to handle Rack devices for undercloud deployments.
Also includes:
1. pep8 corrections for controller_worker.py and a10_config.py
2. Moved rack device read code from __init__ for efficiency 
3. Removed excess calls to "a10_config" object creation


a10_octavia/a10_config.py
```diff
@@ -77,6 +81,48 @@ def __init__(self, config_dir=None, config=None, provider=None):
...
+                    topology = rack_conf["device_list"][i]["topology"]
+                    validation_flag = self.validate(project_id, ip_address, username,
+                                                    password, axapi_version,
+                                                    undercloud, device_name,
+                                                    role, topology)
+
+                    if validation_flag:
+                        vthunder_conf = data_models.VThunder(project_id=project_id,
+                                                             ip_address=ip_address,
+                                                             undercloud=undercloud,
```
omkartelee01: @Hunter, I was looking at your review comment of unpacking dictionary, But I feel, we should make sure that user have created a valid dictionary entry with required parameters. So, I feel this code is necessary to validate the parameters before creating a valid vThunder object

hthompson6: It's simply not the right way to do this check. There is no point in wasting about 80 bytes for variables that'll be declared and used only once and then promptly garbage collected leading to increased overhead.

The validation function already takes in these parameters. So, there is no reason to ensure that they exist in this way when we already have a check ensuring that they exist.

The biggest issue though is that the data model allows for these to not be specified. If they are required, then they need to be positional arguments not keyword args.

hthompson6: If you want better error handling, then make a custom exception for this module that'll catch the TypeError when all the positional args aren't filled or a wrong keyword is specified. You can then have it state a message that "X" wasn't specified in the config.

a10_octavia/a10_config.py
```diff
@@ -145,3 +190,26 @@ def _get_octavia_db_string(self):
     def get(self, key):
         return getattr(self._config, key)
 
+    def validate(self, project_id, ip_address, username, password,
+                 axapi_version, undercloud, device_name, role, topology):
```
hthompson6: Also, `undercloud` isn't used here.

hthompson6: Also, undercloud isn't used here.

a10_octavia/a10_config.py
hthompson6: Also, the issue in the loop still stands of how you're iterating over it. The pythonic way to iterate is with `for-each` loops not `for` loops.

omkartelee01: Still working on the custom error and data model management, will update

PR #22, opened 2019-11-22
STACK-435 : Multi-load balancer configuration on single vthunder

afrin-chakure:
## Highlights

- Added custom network driver called `a10_octavia_neutron` instead of `allowed_address_pairs`.
- Resolved issue related to `ReadTimeOut` error while adding fourth interface to Amphora.

## Non Highlights
Following changes are made:
- Added custom neutron drivers for a10_octavia. Customized neutron methods will reside in `a10_octavia_neutron.py`. Also changed imports from `network_tasks` to `a10_network_tasks` where ever applicable.
- `setup.cfg` - Introduced above driver inside this file.
- `handler_virtual_port.py` : updated default value for `ipinip` to False and updated its handling method.
- `ReadTimeOut Error` - This was coming while setting interface to amphora. Handling is done inside  - `vthunder_tasks.py`
- Lots of pep8 changes including those related to indentation in existing code and line too long.

## Testcase 

Followed below steps
Step1:  CREATE LB
Step2:  CREATE Listener
Step3: CREATE Service group
Step4: Adding members from subnet 1, subnet 2 , subnet 3 , private subnet 

## Result : 

Interfaces attached to Vthunder via UI : 

image

Interfaces attached on Vthunder : 

image

## Closes
[STACK-435](https://a10networks.atlassian.net/browse/STACK-435)


hthompson6:
Will this solution break when it's mixed with rack devices?

afrinc:
@hthompson6 We have removed all the network tasks for rack devices in that flow. So I don't think it will break.

PR #28, opened 2019-12-31
Phase 1 : A10-Health-Manager 

afrin-chakure:
## Non-Highlights 
- Added support for `health-manager` for `a10-octavia`
- Added new service called `a10-health-manager.service`
- support to enable fetching vthunder udp packets 
- added task to switch db roles 

Tested with following configurations in `octavia.conf`

```
bind_port = 5554
bind_ip = 192.168.0.110
heartbeat_interval = 5
heartbeat_key = insecure
heartbeat_timeout = 60
health_check_interval = 3
```

## Closes 
[STACK-448](https://a10networks.atlassian.net/browse/STACK-448)


a10_octavia/db/models.py
```diff
@@ -43,9 +43,11 @@ class VThunder(base_models.BASE):
     compute_id = sa.Column(sa.String(36))
     topology = sa.Column(sa.String(50))
     role = sa.Column(sa.String(50))
+    last_update = sa.Column(sa.DateTime, default=func.now(), nullable=False)
```
omkartelee01: Change attribute name - "last_update" as its very similar to updated_at - can be changed to "last_udp_update"

PR #40, opened 2020-01-28
PEP8 for db module

afrin-chakure:
## Closes
[STACK-638](https://a10networks.atlassian.net/browse/STACK-638)


a10_octavia/db/migration/cli.py
```diff
@@ -27,13 +27,13 @@
 from octavia.db.migration.cli import do_alembic_command, do_check_migration, do_upgrade, no_downgrade, do_stamp, do_revision, add_command_parsers
```
hthompson6: When I ran pep8 it caught these as not being imported....

PR #46, opened 2020-02-26
STACK-806 : Refactored compute tasks

afrin-chakure:
## Closes

STACK-813
STACK-814
STACK-815

### DIFF
--- a/a10_octavia/controller/worker/tasks/a10_compute_tasks.py
+++ b/a10_octavia/controller/worker/tasks/a10_compute_tasks.py
@@ -12,75 +12,40 @@
 #    License for the specific language governing permissions and limitations
 #    under the License.
 
-
 import time
 
-from cryptography import fernet
 from oslo_config import cfg
 from oslo_log import log as logging
-from stevedore import driver as stevedore_driver
-from taskflow import task
 from taskflow.types import failure
 
-from octavia.amphorae.backends.agent import agent_jinja_cfg
 from octavia.common import constants
 from octavia.common import exceptions
-from octavia.common.jinja import user_data_jinja_cfg
-from octavia.common import utils
-from octavia.controller.worker import amphora_rate_limit
-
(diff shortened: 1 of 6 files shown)


a10_octavia/controller/worker/tasks/a10_compute_tasks.py
```diff
@@ -12,76 +12,41 @@
...
+from octavia.controller.worker.tasks.compute_tasks import BaseComputeTask
 
 CONF = cfg.CONF
 LOG = logging.getLogger(__name__)
 
 
-class BaseComputeTask(task.Task):
-    """Base task to load drivers common to the tasks."""
-
+class A10BaseComputeTask(BaseComputeTask):
```
hthompson6: No we don't need this at all. Revert everything back to inheriting from BaseComputeTask

afrin-chakure: Actually while testing it was found that the `__init__()` of `BaseComputeTask`doesn't get called and because of which the `compute_driver` doesn't get initialized.
As a result, the amphora(vThunder) doesn't get booted. Hence we added this bit.

afrin-chakure: This was the error, that was coming. And eventually, it then doesn't find compute_id, as no vthunder is present and hence we get `PlugVIPException`

MicrosoftTeams-image (7)

hthompson6: Issue is being caused by something else then
```
class Parent(object):
    def __init__(self):
        print("Testing")

class Child(Parent):
    pass

c = Child()

hthompson$ python test.py 
Testing
```

omkartelee01: I could not reproduce this error even after multiple times of retries
Reverting the changes we did for this task.

PR #49, opened 2020-03-02
Stack 784 - Revert tasks for handler virtual server

omkartelee01:
## Non-Customer Highlights 
- Delete task for handler virtual server tasks
- Unit tests

## Closes 
[STACK-581](https://a10networks.atlassian.net/browse/STACK-784)

### DIFF
--- a/a10_octavia/controller/worker/tasks/a10_database_tasks.py
+++ b/a10_octavia/controller/worker/tasks/a10_database_tasks.py
@@ -113,7 +113,7 @@ class CreateVThunderEntry(BaseDatabaseTask):
             self.vthunder_repo.delete(
                 db_apis.get_session(), loadbalancer_id=loadbalancer.id)
         except NoResultFound:
-            LOG.info("Could not find vThunder entry to delete.")
+            LOG.error("Failed to delete vThunder entry for load balancer: %s", loadbalancer.id)
 
 
 class DeleteVThunderEntry(BaseDatabaseTask):
--- a/a10_octavia/controller/worker/tasks/a10_database_tasks.py
+++ b/a10_octavia/controller/worker/tasks/a10_database_tasks.py
@@ -107,6 +107,14 @@ class CreateVThunderEntry(BaseDatabaseTask):
 
         LOG.info("Successfully created vthunder entry in database.")
 
+    def revert(self, amphora, loadbalancer, role, status=constants.ACTIVE,
+               *args, **kwargs):
+        try:
+            self.vthunder_repo.delete(
+                db_apis.get_session(), loadbalancer_id=loadbalancer.id)
+        except NoResultFound:
+            LOG.info("Could not find vThunder entry to delete.")
+
 
 class DeleteVThunderEntry(BaseDatabaseTask):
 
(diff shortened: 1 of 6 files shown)


a10_octavia/tests/unit/controller/worker/tasks/test_handler_virtual_server.py
```diff
@@ -1,20 +1,40 @@
...
 
-from a10_octavia.controller.worker.tasks.handler_virtual_server import CreateVitualServerTask, DeleteVitualServerTask
-from a10_octavia.tests.test_base import TestBase
+AMPHORA = o_data_models.Amphora(id=t_constants.MOCK_AMP_ID1)
+VTHUNDER = VThunder()
+LB = o_data_models.LoadBalancer(id=a10constants.MOCK_LOAD_BALANCER_ID, amphorae=[AMPHORA])
 
-USERNAME = "user"
-PASSWORD = "pass"
+class TestCreateVirtualServerTask(base.TestCase):
```
hthompson6: This should be `class TestHanldlerServerTasks(base.Testcase):`

a10_octavia/controller/worker/tasks/handler_virtual_server.py
```diff
@@ -12,91 +12,77 @@
...
         c = self.client_factory(vthunder)
         status = self.set(c.slb.virtual_server.update,
-                          loadbalancer.id,
                           loadbalancer,
                           vthunder)
-        return status
 
-    def revert(self, loadbalancer, vthunder, update_dict):
-        pass
+    def revert(self, loadbalancer, vthunder, *args, **kwargs):
```
hthompson6: We don't need to log this failure here. The failure will propagate up. Only need to revert in the case of orphaned resources.

a10_octavia/tests/unit/controller/worker/tasks/test_handler_virtual_server.py
```diff
@@ -1,20 +1,40 @@
...
 
-class TestCreateVirtualServerTask(TestBase):
     def setUp(self):
+        patcher = patch(
+            'a10_octavia.controller.worker.tasks.common.BaseVThunderTask.client_factory')
+        self.client_factory_mock = patcher.start()
+        self.client_factory_mock.return_value = mock.Mock()
```
hthompson6: Instead we could do

```python
mock_patch = patcher.start()
self.client_mock = mock.Mock()
mock_patch.return_value = self.client_mock
```

Then later on we just have to check ` self.client_mock.slb.virtual_server.delete.assert_called_with(LB.id)`

hthompson6: Better yet done in base class. Since everyone of our task tests uses this

a10_octavia/controller/worker/tasks/a10_database_tasks.py
```diff
@@ -107,6 +107,14 @@ def execute(self, amphora, loadbalancer, role, status=constants.ACTIVE):
 
         LOG.info("Successfully created vthunder entry in database.")
 
+    def revert(self, amphora, loadbalancer, role, status=constants.ACTIVE,
+               *args, **kwargs):
```
hthompson6: Please don't work on things that our out of scope of the PR.

PR #57, opened 2020-03-16
STACK-915 add revert handler for create

sridharsrinivas:


sridharsrinivas:
Forgot to add this.
Testing Done:
Unit Test case  test_revert_pool_create_task
openstack loadbalancer pool create success case
openstack loadbalancer pool create error (with virtual server removed from acos cli).
 openstack loadbalancer pool delete (success and error case).
openstack loadbalancer pool set (success and error(with pool removed from acos cli)).

a10_octavia/controller/worker/tasks/service_group_tasks.py
```diff
@@ -41,11 +41,13 @@ def set(self, set_method, pool, vthunder, update=False):
                        lb_method=lb_method,
                        service_group_templates=service_group_temp,
                        axapi_args=args)
-            LOG.debug("Pool created successfully: %s", pool.id)
+            LOG.debug("Pool created/updated successfully: %s", pool.id)
             return pool
         except Exception as e:
```
hthompson6: Instead of this. Let's change this to simply not catch the error here

a10_octavia/controller/worker/tasks/service_group_tasks.py
```diff
@@ -56,6 +58,15 @@ def execute(self, pool, vthunder):
         axapi_client = self.client_factory(vthunder)
```
hthompson6: Per review comment of set method. Let's instead add a try catch here which handles exceptions coming off the set method. We'll log and then raise.

The last operation of the execute method (that occurs within the try block) will be logging a success

hthompson6: The update method should be altered to match this as well. Though it'll simply catch the exception and log

a10_octavia/tests/unit/controller/worker/tasks/test_handler_service_group.py
```diff
@@ -0,0 +1,33 @@
...
+from a10_octavia.tests.common import a10constants
+from a10_octavia.tests.unit.base import BaseTaskTestCase
+
+VTHUNDER = VThunder()
+POOL = o_data_models.Pool(id=a10constants.MOCK_POOL_ID)
+
+
+class TestHandlerServiceGroupTasks(BaseTaskTestCase):
+
+    def test_revert_pool_create_task(self):
```
hthompson6: You should mock out the task util dependency here too and assert that it was also called with the pool.id

a10_octavia/tests/unit/controller/worker/tasks/test_handler_server.py
```diff
@@ -16,7 +16,7 @@
 from octavia.common import data_models as o_data_models
 from octavia.tests.common import constants as t_constants
 
-from a10_octavia.controller.worker.tasks.handler_server import MemberCreate
+from a10_octavia.controller.worker.tasks.server_tasks import MemberCreate
```
afrin-chakure: good catch

a10_octavia/controller/worker/tasks/service_group_tasks.py
sridharsrinivas: Yes. This is better. Have handled this change.

a10_octavia/controller/worker/tasks/service_group_tasks.py
sridharsrinivas: Did this change.

a10_octavia/tests/unit/controller/worker/tasks/test_handler_service_group.py
sridharsrinivas: Did this change too. Took quite some time though.

PR #58, opened 2020-03-17
PR for STACK-896

hthompson6:
### Closes
[STACK-896](https://a10networks.atlassian.net/browse/STACK-896)


a10_octavia/etc/example_config.py
```diff
@@ -5,43 +5,43 @@
 
 [SLB]
```
afrin-chakure: Can we delete this file completely? @omkartelee01  your comments

a10_octavia/controller/worker/flows/a10_listener_flows.py
```diff
@@ -73,7 +62,7 @@ def get_create_all_listeners_flow(self):
         create_all_listeners_flow.add(database_tasks.ReloadLoadBalancer(
             requires=constants.LOADBALANCER_ID,
             provides=constants.LOADBALANCER))
-        create_all_listeners_flow.add(amphora_driver_tasks.ListenersUpdate(
+        create_all_listeners_flow.add(virtual_port_tasks.ListenersUpdate(
```
omkartelee01: The "get_create_all_listeners_flow" is not being used anywhere, need to remove this flow or start support for this flow. 
Also ListenersUpdate takes 3 args,

a10_octavia/etc/example_config.py
hthompson6: Going to leave it in for this PR since it's targeted towards styling fixes

a10_octavia/controller/worker/flows/a10_listener_flows.py
hthompson6: Going to leave it in for this PR since it's targeted towards styling fixes. We'll remove this during listener flow refactor

PR #68, opened 2020-04-07
STACK-487: Added L3V support

afrin-chakure:
## Non Customer Highlights
- Tested partitions for components like lb, listener, pool and members.

Example :
image

- Note : Currently ignoring sanity checks for `partition` param  in `[rack_vthunder]` as it will be taken care in subsequent PRs and Stories. Note : Currently ignoring sanity checks for partition in `[rack_vthunder]`, as it will be taken care in subsequent PRs and Stories. 

## Closes
[STACK-487](https://a10networks.atlassian.net/browse/STACK-487)


a10_octavia/controller/worker/tasks/decorators.py
```diff
@@ -14,14 +14,33 @@
 
 
 import acos_client
+from acos_client.errors import ACOSException
+from acos_client.errors import Exists
```
hthompson-a10, line 18: One of the Openstack coding guidelines is to not import just a single class.

> Imports¶
Do not import objects, only modules; exceptions are:
imports from migrate package
imports from sqlalchemy package
function imports from i18n module

So instead reformat imports as follows

```python
from acos_client import errors as acos_errors (or a better name)
try:
    import http.client as http_client
except ImportError:
    import httplib as http_client
...
from requests import exceptions as req_exceptions (or a better name)
```

a10_octavia/controller/worker/tasks/decorators.py
```diff
@@ -14,14 +14,33 @@
...
+from requests.exceptions import ReadTimeout
 
 CONF = cfg.CONF
 LOG = logging.getLogger(__name__)
 
 
+def configure_partition(vthunder_client, partition):
+    try:
+        try:
+            vthunder_client.system.partition.create(partition)
```
hthompson-a10, line 36: Making this call every time is a bit much. The only situation in which this should be called is when the load balancer is being created. Let's shift this to a vthunder task called "HandleACOSPartitionChange" or something similar.

This way too we can revert the creation of the partition should an error occur during the flow.

a10_octavia/controller/worker/tasks/decorators.py
```diff
@@ -14,14 +14,33 @@
...
+        except Exists:
+            pass
+        vthunder_client.system.partition.active(partition)
+    except (ConnectionError, ACOSException, BadStatusLine, ReadTimeout):
```
hthompson-a10, line 40: If we get an exception while trying to change the partition we shouldn't ignore it. Any calls made after this will lead to objects being created in the shared partition. We should 100% fail here.

a10_octavia/controller/worker/tasks/decorators.py
afrin-chakure: @hthompson-a10  addressed this in PR https://github.com/a10networks/a10-octavia/pull/70

a10_octavia/controller/worker/tasks/decorators.py
afrin-chakure: Addresses in https://github.com/a10networks/a10-octavia/pull/70

a10_octavia/controller/worker/tasks/decorators.py
afrin-chakure: Addressed in https://github.com/a10networks/a10-octavia/pull/70

afrin-chakure:
This PR had incorrect authors added, hence while amending them, non-required changes commits got included. Hence raised new https://github.com/a10networks/a10-octavia/pull/70

PR #70, opened 2020-04-13
STACK-487: Added L3V support

afrin-chakure-a10:
## Non Customer Highlights
- Tested partitions for components like lb, listener, pool and members for rack devices.
- In case of failure due to wrong configuration under `[rack_vthunder]`, the service stops raising exceptions.

Failure messages : 

Insufficient required params
insuffiencient required params

Invalid IP address
Invalid IP address

duplicate project ids provided
duplicate_project

duplicate ip_address:partition
duplicates_ip_address_partition

Example :
image

## Closes
[STACK-487](https://a10networks.atlassian.net/browse/STACK-487)
[STACK-1055](https://a10networks.atlassian.net/browse/STACK-1055)


a10_octavia/controller/worker/tasks/decorators.py
```diff
@@ -30,6 +43,10 @@ def wrapper(self, *args, **kwargs):
             self.axapi_client = acos_client.Client(vthunder.ip_address, api_ver,
                                                    vthunder.username, vthunder.password,
                                                    timeout=30)
+            if vthunder.topology == constants.ROLE_STANDALONE:
```
hthompson-a10, line 46: If it has a partition in the first place on the config then we know it's a racked device. We don't have support for this in vThunders.

```python
if vthunder.partition != a10constants.SHARED_PARTITION
```

Should suffice as it's not an option for vThunders

a10_octavia/common/config_options.py
```diff
@@ -42,6 +42,9 @@
     cfg.IntOpt('default_axapi_version',
                default=30,
                help=_('VThunder axapi version')),
+    cfg.StrOpt('partition',
```
hthompson-a10, line 45: This instead needs to be relegated solely to the rack device options

a10_octavia/common/data_models.py
```diff
@@ -137,7 +137,7 @@ def __init__(self, id=None, vthunder_id=None, amphora_id=None,
                  password=None, axapi_version=None, undercloud=None,
                  loadbalancer_id=None, project_id=None, compute_id=None,
                  topology="STANDALONE", role="MASTER", last_udp_update=None, status="ACTIVE",
-                 created_at=datetime.utcnow(), updated_at=datetime.utcnow()):
+                 created_at=datetime.utcnow(), updated_at=datetime.utcnow(), partition="shared"):
```
hthompson-a10, line 140: Instead of shared let's keep this as `None` so as to not disrupt vThunders

a10_octavia/controller/worker/tasks/a10_database_tasks.py
```diff
@@ -87,7 +88,8 @@ def execute(self, amphora, loadbalancer, role, status):
             last_udp_update=datetime.utcnow(),
             status=status,
             created_at=datetime.utcnow(),
-            updated_at=datetime.utcnow())
+            updated_at=datetime.utcnow(),
+            partition=partition)
```
hthompson-a10, line 92: Shouldn't need this as we aren't support this for vThunders at this time.

hthompson-a10:
Please change the it to merge into the `feature/l3v_partition` branch instead of `master`

a10_octavia/db/migration/alembic_migrations/versions/ed9c523cf8ee_adding_ha_features.py
```diff
@@ -36,7 +36,8 @@ def upgrade():
         sa.Column('last_udp_update', sa.DateTime(), nullable=True),
         sa.Column('status', sa.String(36), default='ACTIVE', nullable=False),
         sa.Column(u'created_at', sa.DateTime(), nullable=True),
-        sa.Column(u'updated_at', sa.DateTime(), nullable=True)
+        sa.Column(u'updated_at', sa.DateTime(), nullable=True),
+        sa.Column('partition', sa.String(1024), nullable=False)
```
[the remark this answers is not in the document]
afrin-chakure: Its 14 as I see 
image

However need to check on latest vthunder. The rack that I am using has this limit. Once confirmed, will modify this length.

a10_octavia/db/models.py
```diff
@@ -47,6 +47,7 @@ class VThunder(base_models.BASE):
     status = sa.Column('status', sa.String(36), default='ACTIVE', nullable=False)
     created_at = sa.Column(u'created_at', sa.DateTime(), nullable=True)
     updated_at = sa.Column(u'updated_at', sa.DateTime(), nullable=True)
+    partition = sa.Column(sa.String(1024), nullable=False)
```
afrin-chakure, line 50: Will do here also

a10_octavia/controller/worker/flows/a10_load_balancer_flows.py
```diff
@@ -232,6 +232,10 @@ def get_new_lb_networking_subflow(self, topology):
             vthunder_tasks.VThunderComputeConnectivityWait(
                 name=a10constants.MASTER_CONNECTIVITY_WAIT,
                 requires=(a10constants.VTHUNDER, constants.AMPHORA)))
+        if topology == constants.TOPOLOGY_SINGLE:
```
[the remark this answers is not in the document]
afrin-chakure: This flow isn't getting used by rack flow. 
However I inserted this here for single topology flow, which I assumed as part of the story. 

Referring to your  side note: 
Without that topology check, it was throwing error in further task for configuring vrrp saying, the url doesn't exist. Hence added it to avoid it entirely for ACTIVE/STANDBY mode.

Anyways, I think I can remove this task from here, as its not needed.

a10_octavia/controller/worker/tasks/a10_database_tasks.py
afrin-chakure: removing from here.

a10_octavia/controller/worker/tasks/a10_database_tasks.py
```diff
@@ -274,6 +278,7 @@ def execute(self, amphora):
             status="READY",
             created_at=datetime.utcnow(),
             last_udp_update=datetime.utcnow(),
-            updated_at=datetime.utcnow())
+            updated_at=datetime.utcnow(),
+            partition=partition)
```
afrin-chakure, line 282: removing from here as well.

a10_octavia/common/utils.py
```diff
@@ -58,6 +59,8 @@ def convert_to_rack_vthunder_conf(rack_list):
             validation_flag = validate_params(rack_device)
             if validation_flag:
                 rack_device['undercloud'] = True
+                if rack_device['partition'] is None:
```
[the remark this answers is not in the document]
afrin-chakure:
```
if 'partition' not in rack_device or not rack_device['partition']:
       rack_device['partition'] = a10constants.SHARED_PARTITION
```
Handling both empty string case and when attribute itself is not present

a10_octavia/common/config_options.py
afrin-chakure: removing this

a10_octavia/controller/worker/tasks/decorators.py
afrin-chakure: Added it assuming we are doing it for non hardware devices too. Will remove it

a10_octavia/controller/worker/flows/a10_load_balancer_flows.py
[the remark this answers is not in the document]
hthompson-a10: Strange. Our base case is Standalone and then we handle all the backup stuff if topology is Active/Standby.

We can investigate further once this story comes back around again for  non-hardware.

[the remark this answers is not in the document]
afrin-chakure: ok

a10_octavia/common/utils.py
```diff
@@ -40,13 +41,28 @@ def validate_params(rack_info):
...
+            raise exceptions.ValidationException(detail=_(
+                'Invalid IP address given ' + rack_info['ip_address']))
+    raise exceptions.ValidationException(detail=_(
+        'Configuration of `devices` under [RACK_VTHUNDER] is invalid. '
+        'Please check your configuration. The params `project_id`, '
+        '`ip_address`, `username`, `password` and `device_name` cannot be None '))
     return False
 
 
+def check_partition_collision(rack_device):
```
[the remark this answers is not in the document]
afrin-chakure-a10: Since we are calling `validate_params()` before calling this `check_partition_collision()` method, we are making sure - `ip_address` is required along with `project_id`, `username`, `password` and `device_name`. If any one among this is not provided, we are raising exception, with appropriate message.

a10_octavia/common/utils.py
```diff
@@ -40,13 +41,28 @@ def validate_params(rack_info):
...
+    ip_part_set = set()
+    for rack in rack_device.values():
+        candidate = rack.ip_address + rack.partition
```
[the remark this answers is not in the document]
afrin-chakure-a10: Since we are calling `validate_params()` before calling this `check_partition_collision()` method, we are making sure - `ip_address` is required along with `project_id`, `username`, `password` and `device_name`. Similarly if `partition` (not a required param) is provided as None or empty string -> we are setting it to `shared` value

If any one among this is not provided, we are raising exception, with appropriate message.

a10_octavia/common/utils.py
```diff
@@ -58,14 +74,19 @@ def convert_to_rack_vthunder_conf(rack_list):
...
-                            'project ' + rack_device['project_id'])
 
     except KeyError as err:
         LOG.error("Invalid definition of rack device in configuration file."
                   "The Loadbalancer you create shall boot as overcloud."
                   "Check attribute: " + str(err))
-    return rack_dict
+
+    if check_partition_collision(rack_dict):
+        return rack_dict
```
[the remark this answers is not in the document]
afrin-chakure-a10: Its acceptable IMO, as it doesn't matter what we are returning in case of failure, as it won't move forward once exception is raised due to any kind of misconfiguration in Config. The service won't start.

a10_octavia/common/utils.py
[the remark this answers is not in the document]
afrin-chakure-a10: Ok so you want me to store all the duplicated (wrong configs) in a set and then listing them out in the error message.  Modifying the logic accordingly.

a10_octavia/common/utils.py
```diff
@@ -40,32 +40,43 @@ def validate_params(rack_info):
...
-                LOG.warning('Invalid definition of rack device for '
-                            'project ' + rack_device['project_id'])
+    for rack_device in rack_list:
+        validation_flag = validate_params(rack_device)
+        if validation_flag:
+            rack_device['undercloud'] = True
+            if not rack_device.get('partition'):
+                rack_device['partition'] = a10constants.SHARED_PARTITION
+            vthunder_conf = data_models.VThunder(**rack_device)
+            if rack_dict.get(rack_device['project_id']):
```
hthompson-a10, line 71: Let's make this the very first check. This way we don't spend cycles on turning it to a data model or adding partition if it's just going to fail anyway.

a10_octavia/controller/worker/tasks/vthunder_tasks.py
```diff
@@ -333,3 +339,27 @@ def execute(self, master_vrrp_status, backup_vrrp_status):
             return True
         else:
             return False
+
+
+class HandleACOSPartitionChange(task.Task):
```
sridharsrinivas, line 344: Shouldn't this be CreateACOSPartition Task ?

hthompson-a10: It could create it or it could just change to it. We try to not use words like "Create, Update, Delete" unless they exclusively are doing that operation.

One of the ways they've handled this in octavia is by tasks names as "Handle<Object>Delta". So that's being emulated here.

PR #73, opened 2020-04-16
[STACK-1061] Added no-dest-nat check filter

OmkarTelee-A10:
**Issue**

https://a10networks.atlassian.net/browse/STACK-1061

**Test Cases**
We can create 5 types of listener

Enable no_dest_nat in [listener] section of a10-octavia.conf

Create HTTP listener - should get created without no-dest-nat enabled
Create HTTPS listener - should get created without no-dest-nat enabled
Create TERMINATED_HTTP listener - should get created without no-dest-nat enabled

Create TCP listener - should get created with no-dest-nat enabled
Create UDP listener - should get created with no-dest-nat enabled
 
**functional changes**
Add a condition that checks if no-dest-nat is enabled, if protocol is not in TCP or UDP, disable the no-dest-nat argument.

**Non-functional changes**
Added a comment flag so can we can add more filters
As I can see more filters are supposed to be added

### DIFF
--- a/a10_octavia/common/a10constants.py
+++ b/a10_octavia/common/a10constants.py
@@ -42,6 +42,7 @@ SP_OBJ_DICT = {
 }
 HTTP_TYPE = ['HTTP', 'HTTPS']
 PERS_TYPE = ['cookie_persistence', 'src_ip_persistence']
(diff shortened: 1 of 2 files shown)


hthompson-a10:
Will merge after release of v0.1.

PR #75, opened 2020-04-17
STACK-1059 add http templates for HTTPS virtual ports

ssrinivasan-a10:
Testing Done:
1) Add HTTP listener and check for template_http added in conf file gets added.
2) Delete HTTP listener added.
3) Add TERMINATED_HTTPS listener and check for template_http added in conf file gets added.
4) Delete TERMINATED_HTTPS listener added.
5) Add HTTPS listener and check TCP virtual port is created.
6) Delete HTTPS listener added.
7) Add TCP listener and check TCP virtual port is created.
8) Delete TCP listener added.

### DIFF
--- a/a10_octavia/controller/worker/tasks/virtual_port_tasks.py
+++ b/a10_octavia/controller/worker/tasks/virtual_port_tasks.py
@@ -19,6 +19,7 @@ from taskflow import task
 import acos_client.errors as acos_errors
 from octavia.certificates.common.auth.barbican_acl import BarbicanACLAuth
 
+from a10_octavia.common import a10constants
 from a10_octavia.controller.worker.tasks.decorators import axapi_client_decorator
 from a10_octavia.controller.worker.tasks import utils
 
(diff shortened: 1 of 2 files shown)


hthompson-a10:
We should just be able to add listener mappings to here: https://github.com/a10networks/a10-octavia/blob/master/a10_octavia/common/openstack_mappings.py#L15

We can make a mapping call like: https://github.com/a10networks/a10-octavia/blob/b71cb06c142eda10c22e4d9a9acd5adda38b20af/a10_octavia/controller/worker/tasks/health_monitor_tasks.py#L105

a10_octavia/controller/worker/tasks/virtual_port_tasks.py
```diff
@@ -28,6 +29,18 @@
 
 class ListenersParent(object):
 
+    def transform_https_protocols(self, listener):
```
hthompson-a10, line 32: We should just be able to add listener mappings to here: https://github.com/a10networks/a10-octavia/blob/master/a10_octavia/common/openstack_mappings.py#L16

We can make a mapping call like: https://github.com/a10networks/a10-octavia/blob/b71cb06c142eda10c22e4d9a9acd5adda38b20af/a10_octavia/controller/worker/tasks/health_monitor_tasks.py#L105

sridharsrinivas: Done.

PR #80, opened 2020-04-24
[Stack 1060] Switched SOURCE_IP to SOURCE_IP_HASH_ONLY for pools

OmkarTelee-A10:
**Functional change**
As per discussion and a10-docs, figured out SOURCE_IP should map to SOURCE_IP_HASH_ONLY pool lb method.
Switched the mapping in openstack mapping

**Tests** 
Added unit test to make sure SOURCE_IP_HASH_ONLY is passed when a pool created with algorithm SOURCE_IP 

**Files Changed**
a10_octavia/common/openstack_mappings.py
a10_octavia/tests/common/a10constants.py
a10_octavia/tests/unit/controller/worker/tasks/test_service_group_tasks.py


a10_octavia/tests/common/a10constants.py
```diff
@@ -26,3 +26,7 @@
 MOCK_TEMPLATE_NAME = 'mock-template-name-1'
 MOCK_CERT_ACTION = 'import'
 MOCK_CERT_TYPE = 'pem'
+
+SERVICE_GROUP_CONF_SECTION = 'service_group'
+POOL_LB_ALGO_SOURCE_IP = 'SOURCE_IP'
```
hthompson-a10, line 31: Import this instead
https://github.com/openstack/octavia/blob/49036845c118aec11740f18680b779f99113bcfc/octavia/common/constants.py#L120

a10_octavia/tests/common/a10constants.py
```diff
@@ -26,3 +26,7 @@
...
+HTTP_PROTOCOL = 'HTTP'
```
hthompson-a10, line 32: Import this instead
https://github.com/openstack/octavia/blob/49036845c118aec11740f18680b779f99113bcfc/octavia/common/constants.py#L145

a10_octavia/tests/unit/controller/worker/tasks/test_service_group_tasks.py
```diff
@@ -36,9 +41,32 @@ def setUp(self):
...
+
+    def test_create_lb_algorithm_source_ip_hash_only(self):
+        mock_pool = task.PoolCreate()
+        mock_pool.axapi_client = self.client_mock
+        mock_pool.CONF = self.conf
+        POOL = o_data_models.Pool(id=a10constants.MOCK_POOL_ID,
+                                  protocol=a10constants.HTTP_PROTOCOL,
+                                  lb_algorithm=a10constants.POOL_LB_ALGO_SOURCE_IP)
+        mock_pool.execute(POOL, VTHUNDER)
+        self.client_mock.slb.service_group.create.assert_called_with(
```
hthompson-a10, line 66: Since we only care about the lb_method in this test we can instead test with:
```python
self.client_mock.slb.service_group.create.assert_called_with(
    protocol=mock.ANY,
    lb_method=mock_pool.axapi_client.slb.service_group.SOURCE_IP_HASH_ONLY,
    service_group_templates=mock.ANY,
    axapi_args=AXAPI_ARGS)

File: a10_octavia/tests/unit/controller/worker/tasks/test_service_group_tasks.py [hunk 4]
```diff
@@ -36,9 +41,32 @@ def setUp(self):
...
         mock_pool = task.PoolCreate()
         mock_pool.axapi_client = self.client_mock
         mock_pool.revert(POOL, VTHUNDER)
         self.client_mock.slb.service_group.delete.assert_called_with(POOL.id)
+
+    def test_create_lb_algorithm_source_ip_hash_only(self):
+        mock_pool = task.PoolCreate()
+        mock_pool.axapi_client = self.client_mock
+        mock_pool.CONF = self.conf
+        POOL = o_data_models.Pool(id=a10constants.MOCK_POOL_ID,
```
ssrinivasan-a10, line 63: Just a minor comment. Can't we just do an update of global POOL object for keys protocol and lb_algorithm instead of instantiating a new POOL object ?

OmkarTelee-A10: Yes, its question actually, does that affects entire POOL object OR its limited to that function only.

OmkarTelee-A10: As per discussion, we will not create new instance of object if we are not making change in it, So, I am marking this resolved.
We can follow similar format in future as well.

PR #87, opened 2020-05-11
Stack 1209 - HandleVRRPFloatingIPDelta  class with test cases

OmkarTelee-A10:
## Description
Added HandleVRRPFloatingIPDelta task
This task handles, network driver code to add/remove port for VRID floating IP.

**Note**: At this moment, this function DOES NOT take care of configuring VRID via axapi calls. My suggestion is create another task to make that AXAPI call in vthunder_task as this function is already over-populated with logic

Input- VThunder data model with vrrp_floating_ip and vrrp_port_id attributes
output - Returns PORT data model of newly added port, None if no port added
 

## Jira Ticket
[STACK-1209](https://a10networks.atlassian.net/browse/STACK-1209)

## Technical Approach
Added logic to handle dhcp/ip_address values
If missing logic needs to be added 

## Config Changes
None

## Test Cases
1. if vthunder object does not have vrrp_floating_ip, or is None - return None (no port change)
2. if device does not have VRID configured, new port needs to be configured and returned
3, if device have VRID, check if VRID floating IP and vthunder.vrrp_floating_ip are same, No need of change required, return None
4. If device have VRID, check if VRID floating ip and vthunder.vrrp_floating_ip are different, if so, delete old port with vthunder object(using vthunder.vrrp_port_id), create new port
5. if floating ip not in member subnet range, raise error
6. For revert, check if vthunder.vrrp_floating_ip and vthuner.vrrp_port_id exists, if any one of them doesnt exists, do not call delete port.
7.  For revert, check if vthunder.vrrp_floating_ip and vthuner.vrrp_port_id exists, if both of them exists, do call delete port

## Manual Testing
None


a10_octavia/controller/worker/tasks/a10_network_tasks.py
```diff
@@ -658,3 +661,44 @@ def revert(self, result, loadbalancer, amp_data=None, update_dict=None,
         except Exception as e:
             LOG.error('Failed to remove QoS policy: %s from port: %s due '
                       'to error: %s', orig_qos_id, amp_data.vrrp_port_id, e)
+
+
+class HandleVRRPFloatingIPDelta(BaseNetworkTask):
+    """Handle VRRP Floating IP port delta"""
+
+    @axapi_client_decorator
+    def execute(self, vthunder, member):
```
hthompson-a10, line 670:
```python
@axapi_client_decorator
def execute(self, vthunder, member):
    fip_port = None
    floating_ip = vthunder.vrrp_floating_ip

    if floating_ip:
        if floating_ip == 'dhcp':
            fip_port = self.network_driver.create_port(member.subnet_id)
        else:
            subnet_ip, subnet_mask = get_net_info_from_subnet(subnet_id)
            if not a10_utils.check_ip_in_subnet_range(floating_ip, subnet_mask):
                LOG.exception("Invalid VRID floating IP. IP out of subnet range: %s", floating_ip)
                raise

            vrid = self.axapi_client.vrrpa.vrid.get(0)
            vrid_ip = vrid['floating-ip']['ip-address-cfg'][0]['ip-address']
            if floating_ip != vrid_ip:
                fip_port = self.network_driver.create_port(member.subnet_id, fixed_ip=floating_ip)

        if fip_port:
            self.axapi_client.vrrpa.vrid.update(0, fip_port.fixed_ip.ip_address)

    if vthunder.vrrp_port_id:
        self.network_driver.delete_port(vthunder.vrrp_port_id)

    return fip_port
```

a10_octavia/network/drivers/neutron/a10_octavia_neutron.py
```diff
@@ -706,3 +706,25 @@ def wait_for_port_detach(self, amphora):
             except (neutron_client_exceptions.NotFound,
                     neutron_client_exceptions.PortNotFoundClient):
                 pass
+
+    def create_port(self, network_id, fixed_ip=False):
```
hthompson-a10, line 710: Why not use the one already provided upstream?

a10_octavia/tests/unit/controller/worker/tasks/test_network_tasks.py
```diff
@@ -0,0 +1,210 @@
...
+from a10_octavia.tests.common import a10constants
+from a10_octavia.tests.unit import base
+
+MEMBER = o_data_models.Member()
+VTHUNDER = data_models.VThunder()
+SUBNET = o_net_data_models.Subnet()
+PORT = o_net_data_models.Port()
+
+
+class TestNetworkTasks(base.BaseTaskTestCase):
```
hthompson-a10, line 37: Also need a test which covers when no floating ip is specified but a vrrp port id entry exists in the db

a10_octavia/controller/worker/tasks/a10_network_tasks.py
OmkarTelee-A10: This fails in following test case:
if I am updating a member, without any change in it (floating_ip == vrid_ip)
1. New fip_port wont get created
2. The older port will get deleted as 'vthunder.vrrp_port_id' is not None 

I think the change would be, 
if vthunder.vrrp_port_id and fip_port:
        self.network_driver.delete_port(vthunder.vrrp_port_id)

hthompson-a10: Case 1.1: If floating_ip == vrid_ip we don't want to create a new port.
Case 1.2: Good catch. However, we want to handle the case in which no floating_ip is defined in the config yet there is a vrrp_port. We can check `vthunder.vrrp_port_id and floating_ip`

Case 2: If this happens, then the operator didn't configure things properly ahead of time as agreed upon and we should 100% fail. We can't handle everything like we do with standard amphora.

OmkarTelee-A10: there are many bugs with this code, we need to resolve them in pair programming. I am not sure why we are refactoring code, if previous one is covering most of the UTs. Adding remaining test cases is much easier in older code than this refactored one

a10_octavia/tests/unit/controller/worker/tasks/test_network_tasks.py
OmkarTelee-A10: will add UTs after pair programming session.

a10_octavia/controller/worker/tasks/a10_network_tasks.py
hthompson-a10: I'm trying to refactor this code as:
1. It doesn't handle all cases
2. Creates variables that aren't used where they should be (see 'ip_address') with poor naming
3. Is hard to read due to the over use of if statements and returns
4. Doesn't actually update the floating ip on the box itself

All this does in the current state is create or delete a port.

Yes we can devote some time to pair programming to finish this.

a10_octavia/network/drivers/neutron/a10_octavia_neutron.py
hthompson-a10: Ah I had confused it with changes done here: https://github.com/a10networks/a10-octavia/blob/feature/trunking/a10_octavia/network/drivers/neutron/a10_octavia_neutron.py#L104

Which never were merged into master

a10_octavia/tests/unit/controller/worker/tasks/test_network_tasks.py
OmkarTelee-A10: added UT

PR #93, opened 2020-05-14
STACK-1233 Add utility func to patch partial IP octets

ssrinivasan-a10:
## Description
The utility function to patches partial IPv4 suffix with subnet provided.
If the IPV4 suffix contains the complete valid IP then returns the suffix provided.
Otherwise returns the patched IP if it is valid, else returns None.

## Jira Ticket
https://a10networks.atlassian.net/browse/STACK-1233

## Technical Approach
1) For validating the Partial IP provided in the config check the return value of patch_ipv4_suffix(config_partial_ip, '10.0.0.0') to be not None.
2) To patch the partial IP with neutron subnet and validate it, use the return value of patch_ipv4_suffix(config_partial_ip, subnet)

## Test Cases
1) Positive test cases with Partial IP strings '10', '.10', '11.10', '11.0.11.10'.
2) Negative test cases with Partial IP strings '325', 'abc.cef', '11.10.0.11.10', '10.abc.11.10'.

## Manual Testing
1) Tested with STACK-1146 changes.
2) Added "interface_vlan_map" config to rack_vthunder device config.
"interface_vlan_map": {
    "1": {
         "11": {"use_dhcp": False, "ve_ip_address": "20"}, i
         "12": {"ve_ip_address": ".10"}, 
         "13": {"use_dhcp": True}
    }
}
3) Created VIP on VLAN 11 subnet and Member on VLAN 12 subnet.

### DIFF
--- a/a10_octavia/common/utils.py
+++ b/a10_octavia/common/utils.py
@@ -36,30 +36,16 @@ def validate_ipv4(address):
         raise ConfigFileValueError('Invalid IPAddress value given in configuration: ' + address)
 
 
-def patch_ipv4_suffix(suffix, subnet):
-    """Patch the partial IPv4 suffix with subnet provided
-       If suffix contains valid ip returns suffix itself.
-       Returns patched_ip if valid patch else returns None
-    """
-    suffix_octets = suffix.split('.')
-    if suffix_octets[0] == '':
-        suffix_octets.pop(0)
+def validate_partial_ipv4(address):
+    """Validates the partial IPv4 suffix provided"""
+    partial_ip = address.lstrip('.')
+    octets = partial_ip.split('.')
+    for idx in range(4 - len(octets)):
+        octets.insert(0, '0')
 
-    suffix_len = len(suffix_octets)
-    if not 0 < suffix_len <= 4:
-        return None
-
-    prefix_octets = subnet.split('.')
-    idx = 4 - suffix_len
-    while idx > 0:
-        suffix_octets.insert(0, prefix_octets[idx - 1])
-        idx -= 1
-
-    patched_ip = '.'.join(suffix_octets)
-    if netaddr.valid_ipv4(patched_ip, netaddr.core.INET_PTON):
-        return patched_ip
-
-    return None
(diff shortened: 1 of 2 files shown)


a10_octavia/common/utils.py
```diff
@@ -36,6 +36,32 @@ def validate_ipv4(address):
...
+    """
+    suffix_octets = suffix.split('.')
+    if suffix_octets[0] == '':
+        suffix_octets.pop(0)
+
+    suffix_len = len(suffix_octets)
+    if not 0 < suffix_len <= 4:
+        return None
+
+    prefix_octets = subnet.split('.')
```
hthompson-a10, line 52: We can't do this with the subnet at the config stage since we don't know the subnet or the netmask. We are going to handle the subnet side of things later.

ssrinivasan-a10: At the config level we provide a fixed string like "10.0.0.0" like I specified in Technical Approach point 1. I just named the string subnet and we use the prefix only. I can rename the argument to prefix.
At the runtime when we get the cidr subnet, we need to pass that as prefix to the function.

hthompson-a10: We don't want to have to have them specify the exact subnet ip in the config file. That defeats the whole purpose of this feature. They could easily create the LB which will cause this information to be stored long before they decide where they want the members to be created.

The runtime check doesn't really get us anywhere. I could pass `15.8.9` and subnet `12.13.0.0` and it'll spit out `12.15.8.9` but the netmask is `255.255.240.0` it's invalid so we just did a bunch of work for nothing. Better to just set it to `0.15.8.9`. Then we can just confirm it's in range and do a bitwise & to set the network ip. This is faster and less LOC than what is being done here.

All this code should do is transform `5`, `.5`, `.5.2.1` to `0.0.0.5`, `0.0.0.5`, `0.5.2.1`. Then check that is is an actually valid IP so we don't have a weird `.777` being passed in or something when they meant `.77`

ssrinivasan-a10: Did the changes.

PR #96, opened 2020-05-19
STACK-1063 Implement VE and VLAN for Hardware Devices (DHCP only)

afrin-chakure-a10:
## Description
- Added feature to enable communication over vlans with traffic isolation for Hardware devices.

## Jira Ticket
[STACK-1063](https://a10networks.atlassian.net/browse/STACK-1063)

## Technical Approach
- Added Config option - `network_type` for user to specify `vlan` or `flat` hierarchy
- Under `[hardware_thunder]`, user can add `interface_vlan_map` to specify configuring ve via dhcp or not. Added column in `vthunders` table and `VThunder` model.
- Added validations for `interface_vlan_map` in `utils.py`
- Separated flat and vlan subflows.
- Added method in `utils` to get network driver.
- Added task to create and delete TagEthernet for both loadbalancer and member.
- Added methods to create, get and delete neutron ports.
- Refactored `config_types.py` 

## Config Changes
<pre>
<b>[a10_global]</b>
network_type = 'vlan'

<b>[hardware_thunder]</b>
devices=[
 {
 "project_id": "fake_uuid",
 "username" : "username",
 "password" : "password",
 "ip_address" : "10.43.12.137",
 "device_name" : "rack_1",
 "interface_vlan_map":{
                 {"1":
                       {"11": {"use_dhcp": True}, 
                         "12":{"use_dhcp": True}
                        }
                 }
       }
 }]
</pre>

## Test Cases
- Updated UTs to handle positive and negative test cases.

## Manual Testing
1) Add 1 VIP and 2 Members in 3 different VLANs and send traffic
2) Delete the Members and VIP one by one and check the VLANs and VE deleted
3) Add VIP and Member from same VLAN and another member in different VLAN send traffic
4) After above case Delete the Member on VIP VLAN and send traffic
5) Change the VE IPs in a10-octavia.conf, restart a10-controller worker and add a member. Check the VE IPs changed accordingly in thunder. Send traffic.
6) Again change VE IPs in conf, restart controller worker and disable a member. Check the VE IPs changed accordingly in thunder. Send traffic.
7) With network_type flat add 1 VIP and 2 Members and send traffic.
8) With above setup remove 1 Member and send traffic.
9) With above setup remove the remaining entities and check configuration.

PR #97, opened 2020-05-21
STACK-1253 and STACK-1244 - Refactor and unison a10-octavia-neutron driver

OmkarTelee-A10:
## Description
Refactored neutron driver
Added create and delete port to a10-octavia-neutron driver
Merged together code for trunking + VRID + auto ve/vlan branches - code credits - @hthompson-a10  @ssrinivasan-a10 

## Jira Ticket
[STACK-1253](https://a10networks.atlassian.net/browse/STACK-1253)
[STACK-1244](https://a10networks.atlassian.net/browse/STACK-1244)

## Technical Approach
Used allowedaddresspair as parent driver to write custom a10-octavia-neutron driver so that we can enhance the functionalities supported by the neutron driver.
Added create and delete port functions, used feature/trunking branch to add trunking solutions to the master.

## Test Cases
- Booted a vThunder with normal approach
- Booted a vThunder with rack approach
- Created a port for VRID member creation 
- Deleted a port for VRID member deletion

## Manual Testing
- Create a LB, check network driver was called
- create a LB on rack, check network driver was not called
- create a member with VRID, check create port called
- change VRID, check delete port was called

### DIFF
--- a/a10_octavia/common/utils.py
+++ b/a10_octavia/common/utils.py
@@ -21,11 +21,12 @@ import netaddr
 import socket
 import struct
 
-from oslo_config.cfg import ConfigFileValueError
+from oslo_config import cfg
 from oslo_log import log as logging
 
 from keystoneclient.v3 import client as keystone_client
 from octavia.common import keystone
+from stevedore import driver as stevedore_driver
 
 from a10_octavia.common import a10constants
 from a10_octavia.common import data_models
@@ -36,7 +37,8 @@ LOG = logging.getLogger(__name__)
 def validate_ipv4(address):
     """Validates for IP4 address format"""
     if not netaddr.valid_ipv4(address, netaddr.core.INET_PTON):
-        raise ConfigFileValueError('Invalid IPAddress value given in configuration: ' + address)
+        raise cfg.ConfigFileValueError(
+            'Invalid IPAddress value given in configuration: ' + address)
 
 
 def validate_partial_ipv4(address):
@@ -47,8 +49,8 @@ def validate_partial_ipv4(address):
         octets.insert(0, '0')
 
     if not netaddr.valid_ipv4('.'.join(octets), netaddr.core.INET_PTON):
-        raise ConfigFileValueError('Invalid partial IPAddress value given in configuration: '
-                                   + address)
+        raise cfg.ConfigFileValueError('Invalid partial IPAddress value given in configuration: '
+                                       + address)
 
 
 def validate_partition(hardware_device):
@@ -71,9 +73,9 @@ def validate_params(hardware_info):
             validate_ipv4(hardware_info['ip_address'])
             hardware_info = validate_partition(hardware_info)
             return hardware_info
-    raise ConfigFileValueError('Please check your configuration. The params `project_id`, '
-                               '`ip_address`, `username`, `password` and `device_name` '
-                               'under [hardware_thunder] section cannot be None ')
+    raise cfg.ConfigFileValueError('Please check your configuration. The params `project_id`, '
+                                   '`ip_address`, `username`, `password` and `device_name` '
+                                   'under [hardware_thunder] section cannot be None ')
 
 
 def check_duplicate_entries(hardware_dict):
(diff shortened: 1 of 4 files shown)


a10_octavia/common/utils.py
```diff
@@ -36,7 +37,8 @@
 def validate_ipv4(address):
     """Validates for IP4 address format"""
     if not netaddr.valid_ipv4(address, netaddr.core.INET_PTON):
-        raise ConfigFileValueError('Invalid IPAddress value given in configuration: ' + address)
+        raise cfg.ConfigFileValueError(
+            'Invalid IPAddress value given in configuration: ' + address)
```
hthompson-a10, line 41: Safer to use `.format()`

a10_octavia/network/drivers/neutron/a10_octavia_neutron.py
```diff
@@ -12,697 +12,189 @@
...
 
-from octavia.common import constants
-from octavia.common import data_models
-from octavia.common import exceptions
-from octavia.i18n import _
-from octavia.network import base
+from neutronclient.common import exceptions as neutron_client_exceptions
 from octavia.network import data_models as n_data_models
-from octavia.network.drivers.neutron import base as neutron_base
+from octavia.network.drivers.neutron.allowed_address_pairs import AllowedAddressPairsDriver
```
hthompson-a10, line 22: Have to import this as a module based on styling guide

a10_octavia/network/drivers/neutron/a10_octavia_neutron.py
```diff
@@ -12,697 +12,189 @@
...
-        except (neutron_client_exceptions.NotFound,
-                neutron_client_exceptions.PortNotFoundClient):
-            pass
-        except Exception as e:
-            LOG.error('Failed to delete port.  Resources may still be in '
-                      'use for port: %(port)s due to error: %s(except)s',
-                      {'port': amphora.vrrp_port_id, 'except': e})
 
-    def unplug_vip(self, load_balancer, vip):
+    def create_ve_port(self, ip, subnet):
```
hthompson-a10, line 164: This isn't required. Important piece is `fixed_ips` being set in the code below; however, we can do that just by modifying the `fixed_ip` argument to be `fixed_ips` in method signature of `create_port`. It'll accept `{'ip_address': ip, 'subnet_id': subnet.id}` mappings.

a10_octavia/network/drivers/neutron/a10_octavia_neutron.py
```diff
@@ -12,697 +12,189 @@
...
-            try:
-                self.neutron_client.update_port(port.id,
-                                                {'port': {'dns_name': ''}})
-
-            except (neutron_client_exceptions.NotFound,
-                    neutron_client_exceptions.PortNotFoundClient):
-                raise base.PortNotFound()
-
-    def plug_port(self, amphora, port):
+    def get_ve_port_id(self, ip):
```
hthompson-a10, line 183: We should just have a default `get_port` that maps it to the datamodel. I believe this utility already exists though in the upstream octavia code base.

afrin-chakure-a10: Yes `get_port` does exist, but I guess, taking this changes from #96 which get port details by Ip address, needs this method. 
Hence changing the method name as `get_port_by_name` and also making it return `network_models.Port(id=port['id'])`

hthompson-a10: Actually this isn't even required. The only reason this was being done is because the `create_ve_port` wasn't returning the port itself. We should never have to get one of the ports we reserve by IP.

a10_octavia/network/drivers/neutron/a10_octavia_neutron.py
```diff
@@ -12,697 +12,169 @@
...
-        except Exception:
-            message = _('Error unplugging VIP. Could not clear '
-                        'allowed address pairs from port '
-                        '{port_id}.').format(port_id=vip.port_id)
+            message = "Error deleting port: {0}".format(port_id)
             LOG.exception(message)
-            raise base.UnplugVIPException(message)
 
-        # Delete the VRRP port if we created it
+    def get_port_by_ip(self, ip):
```
hthompson-a10, line 163: This isn't required. The only reason this was done was because the `create_ve_port` was not returning the port it was created.

PR #118, opened 2020-06-25
STACK-1386 : Added exceptions when invalid value is provided for interface_vlan_map

afrin-chakure-a10:
## Description
- Severity Level : High
- Required: Adding exceptions for invalid datatype for `interface_num`, `vlan_id` and `use_dhcp`.

## Jira Ticket
[STACK-1386](https://a10networks.atlassian.net/browse/STACK-1386)

## Technical Approach
- Raise custom exceptions with proper error message in case of invalid value provided for  `interface_num`, `vlan_id` and `use_dhcp`.
- Also included UT test case fix for `utils.py` module.
- Fixed pep8 fixes travis failure.

## Config Changes
None

## Test Cases
- `interface_num` should be Integer value. Otherwise throw Config error.
- `vlan_id` should be Integer value. Otherwise throw Config error.
- `use_dhcp` should be bool value. Otherwise throw Config error.

Screenshots
interface_num
vlan_id
use_dhcp

## Manual Testing
- Change `a10-octavia.conf` and restart `a10-controller-worker.service`.

### DIFF
--- a/a10_octavia/common/utils.py
+++ b/a10_octavia/common/utils.py
@@ -221,7 +221,7 @@ def convert_interface_to_data_model(interface_obj):
         if vlan_id in interface_dm.tags:
             raise exceptions.DuplicateVlanTagsConfigError(interface_num, vlan_id)
         if vlan_map.get('use_dhcp'):
-            if not type(vlan_map.get('use_dhcp')) == bool:
+            if not vlan_map.get('use_dhcp') in ("True", "False"):
                 raise exceptions.InvalidUseDhcpConfigError(vlan_map.get('use_dhcp'))
             if vlan_map.get('ve_ip'):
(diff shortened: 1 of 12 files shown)


a10_octavia/common/utils.py
```diff
@@ -191,15 +191,21 @@ def convert_interface_to_data_model(interface_obj):
     interface_dm = data_models.Interface()
     if not interface_num:
         raise exceptions.MissingInterfaceNumConfigError()
+    if not isinstance(interface_num, int):
```
hthompson-a10, line 194: In all the below cases, it's highly unlikely that these are inheriting from some Integer superclass. Therefore, `type()` is a more appropriate check

afrin-chakure-a10: ok

PR #121, opened 2020-07-01
STACK-1314: Add write mem to end of flows

hthompson-a10:
## Description
Writes device memory at the end of each flow

## Jira Ticket
[STACK-1314](https://a10networks.atlassian.net/browse/STACK-1314)

## Technical Approach
- Required: Overview of approach taken to solve the problem
- Optional: List of tasks completed w/ links 

## Test Cases
- Write to the shared partition calls client
- Write to a given partition calls client with partition name

## Manual Testing
- Execute against each flow
- Check that taking down the active doesn't affect write memory of the standby
- Check that taking down the standby doesn't affect write memory of the active
- Ensure that in active-standby mode memory is saved on both devices


hthompson-a10:
Rebased against master instead of feature branch. Force pushed before noticing. PR #123 has replaced this PR.

PR #125, opened 2020-07-06
[BUG][STACK-1322]: HM isn't removed from Thunder if delete a pool from Octavia

afrin-chakure-a10:
## Description
- Severity Level (Low, High, Critical)
- Health monitor configuration was not getting removed from vThunder when we delete its associated Pool from openstack, even though, db entry for `health_monitor` gets removed.

## Jira Ticket
[STACK-1322](https://a10networks.atlassian.net/browse/STACK-1322)

## Technical Approach
- Invoked `delete_health_monitor` flow inside `delete_pool` flow, if there exist an hm associated with that pool.
- Since the flow for pool, member and hm is same for hardware_thunder and Vthunder flows, it will work in similar way.

## Config Changes
None

## Test Cases
- When pool is deleted via openstack cli command, the `health_monitor` associated with it are also deleted from vthunder config.

## Manual Testing
- `openstack loadbalancer pool delete pool1`
- Check vThunder, along with `pool`, associated `health_monitor` also gets deleted from `show running-config`.

_**Please Note :**_ 

Because of db task of `DecrementHealthMonitorQuota` or similary `DecrementMemberQuota` in the hm flow or member_flow respectively, warning is issued 
_More details about these tasks in Octavia as below:_
https://github.com/openstack/octavia/blob/1ea2fb18c076049515354ca37d650a5862302a9f/octavia/controller/worker/tasks/database_tasks.py#L2482
https://github.com/openstack/octavia/blob/73fbc05386b512aa1dd86a0ed6e8455cc6b8dc7f/octavia/db/repositories.py#L508

image

Which is understandable, in case when `hm/member` gets deleted via `pool` delete.
Checked Quotas table, and they give correct numbers used: before, during and after invoking cli commands.


afrin-chakure-a10:
Will update this with a better approach without calling the entire flow.
Closing this PR for now, instead will raise a new PR

PR #126, opened 2020-07-06
Stack 1402 - VE/VLAN branch not working with single device

OmkarTelee-A10:
## Description

Issue Description:
The aVCS config is not working with single device setup due to missing network device map check.

Severity Level: Critical

## Jira Ticket
https://a10networks.atlassian.net/browse/STACK-1402

## Technical Approach
Added a condition that checks that device network map is provided.

## Test Cases
Environment in which both single and clustered setup co-exists.
Project1 must be aVCS configured and project2 must be a single rack device.
Create LB in both projects, both gets created successfully.

## Manual Testing
Created a project, associated proper users.
In config added a device as rack device for project.
For other project, in config added a aVCS cluster floating IP and configuration.
Booted LBs in both project.


afrin-chakure-a10:
Looks good to me 👍

PR #133, opened 2020-07-14
[DO NOT MERGE] Stack 1429 - Review and correct flow & tasks for Health Monitor

rahulkatre-a10:
## Description
The creation and deletion of a health monitor operations were not atomic. Also, Checked if the act of creation and deletion causes any of load balancer, the listener. pool in the pending state. 

If Bug Fix:
- Severity Level: High
- Issue Description: The creation and deletion of a health monitor operations were not atomic. Also, Checked if the act of creation and deletion causes any of load balancer, the listener. pool in pending state. 

## Jira Ticket
https://a10networks.atlassian.net/browse/STACK-1429

## Technical Approach
- Added revert for creating and deleting health monitor.

## Manual Testing
- Tested creation of health monitor for pool.
- Tested creation flow by raising an exception on the creation of HM, and checked whole tree(load balancer- listener-pool ) to validate if those objects in Active or pending state.
- Tested if revert specific to create hm task getting executed or not.
- Tested deletion of health monitor flow.
- Tested deletion flow by raising an exception on the deletion of HM, and checked whole tree(load balancer- listener-pool ) to validate if those objects in Active or pending state.
- Tested if revert specific to delete hm task getting executed or not.

### DIFF
--- a/a10_octavia/common/exceptions.py
+++ b/a10_octavia/common/exceptions.py
@@ -129,7 +129,3 @@ class MissingMgmtIpConfigError(cfg.ConfigFileValueError):
         msg = ('Missing `mgmt_ip_address` for vcs device with id {0}. ' +
                'Please provide management IP address').format(vcs_device_id)
         super(MissingMgmtIpConfigError, self).__init__(msg=msg)
-
-
-class GenericFlowException(exceptions.OctaviaException):
-    pass
--- a/a10_octavia/common/exceptions.py
+++ b/a10_octavia/common/exceptions.py
@@ -129,3 +129,7 @@ class MissingMgmtIpConfigError(cfg.ConfigFileValueError):
         msg = ('Missing `mgmt_ip_address` for vcs device with id {0}. ' +
                'Please provide management IP address').format(vcs_device_id)
         super(MissingMgmtIpConfigError, self).__init__(msg=msg)
+
+
+class GenericFlowException(exceptions.OctaviaException):
+    pass
--- a/a10_octavia/controller/worker/tasks/health_monitor_tasks.py
+++ b/a10_octavia/controller/worker/tasks/health_monitor_tasks.py
@@ -42,7 +42,7 @@ class CreateAndAssociateHealthMonitor(task.Task):
         args = utils.meta(health_mon, 'hm', {})
 
         try:
-            self.axapi_client.slb.hm.create(health_mon.id[0:5],
+            self.axapi_client.slb.hm.create(health_mon.id,
                                             openstack_mappings.hm_type(self.axapi_client,
                                                                        health_mon.type),
                                             health_mon.delay, health_mon.timeout,
@@ -56,7 +56,7 @@ class CreateAndAssociateHealthMonitor(task.Task):
 
         try:
             self.axapi_client.slb.service_group.update(health_mon.pool_id,
-                                                       health_monitor=health_mon.id[0:5],
+                                                       health_monitor=health_mon.id,
                                                        health_check_disable=0)
             LOG.debug("Health Monitor %s is associated to pool %s successfully.",
                       health_mon.id, health_mon.pool_id)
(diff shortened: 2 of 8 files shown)


a10_octavia/controller/worker/tasks/health_monitor_tasks.py
```diff
@@ -77,12 +87,26 @@ def execute(self, health_mon, vthunder):
             LOG.debug("Health Monitor %s is dissociated from pool %s successfully.",
                       health_mon.id, health_mon.pool_id)
         except Exception as e:
-            LOG.warning("Failed to dissociate Health Monitor from pool: %s", str(e))
+            msg = str(e)
+            LOG.exception("Failed to dissociate Health Monitor from pool: %s", msg)
+            raise GenericFlowException(msg)
```
hthompson-a10, line 92: Catching all and then throwing a GenericFlowException will obscure any real errors. There are two failure types expected here:
1. `ACOSException` from the `acos_client` library
2. `ConnectTimeout` from the `requests` library

Those should be *can* be caught if we desire to raise a log message that isn't a stacktrace. We should catch both of those and re-raise the errors with the `id` of the health monitor.

There is no need to catch all here since we know the errors that can be produced.

a10_octavia/common/exceptions.py
```diff
@@ -129,3 +129,7 @@ def __init__(self, vcs_device_id):
         msg = ('Missing `mgmt_ip_address` for vcs device with id {0}. ' +
                'Please provide management IP address').format(vcs_device_id)
         super(MissingMgmtIpConfigError, self).__init__(msg=msg)
+
+
+class GenericFlowException(exceptions.OctaviaException):
```
hthompson-a10, line 134: This is not required for this refactor. Please see other comment.

a10_octavia/controller/worker/tasks/health_monitor_tasks.py
```diff
@@ -76,13 +87,16 @@ def execute(self, health_mon, vthunder):
...
                       health_mon.id, health_mon.pool_id)
-        except Exception as e:
-            LOG.warning("Failed to dissociate Health Monitor from pool: %s", str(e))
+        except (acos_errors.ACOSException, ConnectionError) as e:
+            LOG.exception("Failed to dissociate pool %s from health monitor: %s", health_mon.pool_id, health_mon.id)
+            raise e
+
         try:
             self.axapi_client.slb.hm.delete(health_mon.id[0:5])
             LOG.debug("Health Monitor deleted successfully: %s", health_mon.id)
```
afrin-chakure-a10, line 105: Please change this message to : `"Successfully deleted health monitor: %s"`

a10_octavia/controller/worker/tasks/health_monitor_tasks.py
```diff
@@ -109,5 +123,6 @@ def execute(self, listeners, health_mon, vthunder, update_dict):
                 method=method, url=url, expect_code=expect_code,
                 port=listeners[0].protocol_port, axapi_args=args)
             LOG.debug("Health Monitor updated successfully: %s", health_mon.id)
```
afrin-chakure-a10, line 134: Please change this message to : `"Successfully updated health monitor: %s"`

a10_octavia/controller/worker/tasks/health_monitor_tasks.py
```diff
@@ -109,5 +123,6 @@ def execute(self, listeners, health_mon, vthunder, update_dict):
...
-        except Exception as e:
-            LOG.exception("Failed to update Health Monitor: %s", str(e))
+        except (acos_errors.ACOSException, ConnectionError) as e:
+            LOG.exception("Failed to update Health Monitor: %s", health_mon.id)
```
afrin-chakure-a10, line 136: small case `health monitor` in error message.

PR #154, opened 2020-07-29
[Stack 1486] Raise ProviderUnsupportedOptionError for PROXY pool

OmkarTelee-A10:
## Description
User getting unwanted error when he/she is trying to create a pool with PROXY protocol.

## Severity Level 
 High

## Jira Ticket
https://a10networks.atlassian.net/browse/STACK-1486

## Technical Approach
Add a check before creating a pool to check protocol type.
If type is proxy, raising appropriate error

## Test Cases
Create a pool without proxy protocol
Create a pool with proxy protocol make sure correct error is thrown

## Manual Testing
` ERROR oslo_messaging.rpc.server [-] Exception during message handling: ProviderUnsupportedOptionError: Provider 'A10' does not support a requested option: A pool with protocol PROXY is not supported by A10 provider.`

### DIFF
--- a/a10_octavia/controller/worker/tasks/service_group_tasks.py
+++ b/a10_octavia/controller/worker/tasks/service_group_tasks.py
@@ -17,6 +17,9 @@ from oslo_log import log as logging
 from requests.exceptions import ConnectionError
 from taskflow import task
 
+from octavia.common import constants
+from octavia.common import exceptions
+
 import acos_client.errors as acos_errors
 
 from a10_octavia.common import openstack_mappings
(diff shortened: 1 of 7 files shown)


a10_octavia/controller/worker/tasks/service_group_tasks.py
```diff
@@ -51,6 +56,11 @@ class PoolCreate(PoolParent, task.Task):
     @axapi_client_decorator
     def execute(self, pool, vthunder):
         try:
+            if pool.protocol == constants.PROTOCOL_PROXY:
+                raise exceptions.ProviderUnsupportedOptionError(
```
hthompson-a10, line 60: Please include some information about the pool and vthunder. Ideally their ids. Operators will be using automation tools to build things so tracking this error down will be difficult.

OmkarTelee-A10: added pool id in log message. I cant accept it as a special variable, so embeded in same

PR #156, opened 2020-08-05
[BUG][STACK-1393]- Assigning None to template name fails listener creation.

rahulkatre-a10:
## Description
Assigning None to the template name(i.e, template_* = None) considers None as the name of the template, and since such a template does not exist on vthunder ie gives error while creating a listener.

Severity Level: High

## Jira Ticket
https://a10networks.atlassian.net/browse/STACK-1393

## Technical Approach
- Checked for none string if so replace that string with None object.

## Manual Testing
- Create a template on vThunder and created a listener with this template.
- Assigned None to "template_virtual_port" and checked whether the creation of listener fails.
- Assigned None to "template_tcp" and checked whether the creation of the listener fails.
- Assigned None to "template_policy" and checked whether the creation of the listener fails.
- Assigned None to "template_http" and checked whether the creation of the listener fails.

### DIFF
--- a/a10_octavia/common/a10constants.py
+++ b/a10_octavia/common/a10constants.py
@@ -101,4 +101,3 @@ DELETE_MEMBERS_SUBFLOW_WITH_POOL_DELETE_FLOW = 'delete-members-subflow-with-pool
 HANDLE_VRID_MEMBER_SUBFLOW = 'handle-vrid-member-subflow'
 SPARE_VTHUNDER_CREATE = 'spare-vthunder-create'
 LB_TO_VTHUNDER_SUBFLOW = 'lb-to-vthunder-subflow'
-NONE_STR = ["None"]
--- a/a10_octavia/common/a10constants.py
+++ b/a10_octavia/common/a10constants.py
@@ -101,3 +101,4 @@ DELETE_MEMBERS_SUBFLOW_WITH_POOL_DELETE_FLOW = 'delete-members-subflow-with-pool
 HANDLE_VRID_MEMBER_SUBFLOW = 'handle-vrid-member-subflow'
 SPARE_VTHUNDER_CREATE = 'spare-vthunder-create'
 LB_TO_VTHUNDER_SUBFLOW = 'lb-to-vthunder-subflow'
+NONE_STR = ["None"]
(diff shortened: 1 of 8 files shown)


a10_octavia/common/a10constants.py
```diff
@@ -101,3 +101,4 @@
 HANDLE_VRID_MEMBER_SUBFLOW = 'handle-vrid-member-subflow'
 SPARE_VTHUNDER_CREATE = 'spare-vthunder-create'
 LB_TO_VTHUNDER_SUBFLOW = 'lb-to-vthunder-subflow'
+NONE_STR = ["None"]
```
hthompson-a10, line 104: There isn't a real need to make a constant for this. It isn't widely used nor is it continually subject to change like flow names are.

a10_octavia/controller/worker/tasks/virtual_port_tasks.py
```diff
@@ -41,7 +41,8 @@ def set(self, set_method, loadbalancer, listener, ssl_template=None):
 
         virtual_port_templates = {}
         template_virtual_port = CONF.listener.template_virtual_port
-        virtual_port_templates['template-virtual-port'] = template_virtual_port
+        virtual_port_templates['template-virtual-port'] = utils.get_template_name(
```
hthompson-a10, line 44: There isn't a need to go so far as a unique utility function. We are already defining a variable here so we might as well use it for real.

For line length concerns we can cut down `virtual_port_templates` to `vport_templates`

```python
template_vport = CONF.listener.template_virtual_port
vport_templates['template-virtual-port'] = template_vport if template_vport != "None" else None
```

Later we can just use name the template vars better for example

```python
template_http = CONF.listener.template_http
vport_templates['template-http'] = template_http if template_http != "None" else None
```

a10_octavia/common/a10constants.py
rahulkatre-a10: removed it.

a10_octavia/controller/worker/tasks/virtual_port_tasks.py
rahulkatre-a10: Not used ternary operator, added seperate condition.

rahulkatre-a10:
> Please squash and reword commit history.
> 
> Commits should be in the format of "Did X to resolve Y" though I'll accept "Did X". These should be written in past tense and operate as a description of work done beyond "Added condition".
> 
> See: #136

Squashed commits and updated the commit message.

PR #163, opened 2020-08-20
[STACK-1457]: Deleting a health monitor from pool sets pool to health-check-disable instead of just removing the hm association.

rahulkatre-a10:
## Description
When deleting a health monitor, health_check_disable will also get set. As per expectation, the reference of health monitor should be removed from the service group and health_check_disable attribute should be unaltered.

**Before fix behavior:**
`When attached HM is deleted        

slb service-group 8f60340f-e034-403d-83d3-006afff157e9 tcp  

  health-check-disable  

  member 420735bc-2a96-4b45-9be7-32b22b82f476 80  `

**After fix behavior:**
`openstack loadbalancer healthmonitor create --delay 10 --timeout 5  --max-retries 4 --type HTTP --http-method HEAD --url-path "/abc" --name hm-2 --expected-codes 208 pool-psi 

slb service-group 8d717f7d-14f6-4794-b3ff-405dc9fafc7e tcp  
  health-check 959c7349-d9e9-4378-9e9e-4aea96fbfe99  
!    

openstack loadbalancer healthmonitor delete hm-2 

slb service-group 8d717f7d-14f6-4794-b3ff-405dc9fafc7e tcp  
!   `

If Bug Fix:
High

## Jira Ticket
https://a10networks.atlassian.net/browse/STACK-1457

## Technical Approach

1. Initially checked if removing or setting false or None health_check_disable works, axapi giving an error.
2. Compared service group create API payload with service group update API payload.
3. Statically added the health-check parameter to API payload while updating service group object, worked.
4. Integrated changes with the appropriate condition.

## Manual Testing
Tested creation of service group object.
Tested update of service group object.
Tested delete service group object.
Tested the creation of health monitor.
Tested deletion of health monitor.

### DIFF
--- a/a10_octavia/controller/worker/tasks/health_monitor_tasks.py
+++ b/a10_octavia/controller/worker/tasks/health_monitor_tasks.py
@@ -88,8 +88,7 @@ class DeleteHealthMonitor(task.Task):
     def execute(self, health_mon, vthunder):
         try:
             self.axapi_client.slb.service_group.update(health_mon.pool_id,
-                                                       health_monitor="",
-                                                       health_check_disable=True)
+                                                       health_check=None)
             LOG.debug("Successfully dissociated health monitor %s from pool %s",
                       health_mon.id, health_mon.pool_id)
         except (acos_errors.ACOSException, ConnectionError) as e:


afrin-chakure-a10 (approved):
👍 Great

PR #175, opened 2020-09-08
[BUG][STACK-1580] : LB should not be created in parent partition with use_parent_partition False and hmt enabled.

afrin-chakure-a10:
## Description
- Severity Level : **High**
- When we create LB with `use_parent_partition` flag as `False` and `hierarchical_multitenancy` (hmt) as `enable` n `a10-octavia.conf`, it should get created in project partition/ shared partition and not in parent partition.

## Jira Ticket
https://a10networks.atlassian.net/browse/STACK-1580

## Technical Approach
- By default, `hierarchical_multitenancy` (hmt) is `disable`, if not set.
- Changed if condition to be more specific, to check if hmt is set to `enable` and `use_parent_partition` as True, then only Parent partition is to be used. Otherwise for all other conditions, partition used will be either `partition_name` or `shared` (if `partition_name` is not mentioned).

_**Update**_: 
- Added log error for project having no parent or default as parent when all parent partition settings are enabled.

image

- Also added `revert` for `CreateRackVthunderEntry`, to delete vthunder entry during reverts of taskflow.

## Config Changes
None

## Test Cases

SR No | hierarchical_multitenancy | use_parent_partition | partition_name | Result (Partition used) | LOGS/Warning
-- | -- | -- | -- | -- | --
  |   |   |   |   |  
1 | disable | FALSE | partitionA | partitionA |  -
2 | disable | FALSE | - | shared |  -
3 | disable | TRUE | partitionA | partitionA | *Warning
4 | disable | TRUE | - | shared | *Warning
5 | enable | FALSE | partitionA | existing project id [:14] |  -
6 | enable | FALSE | - | existing project id [:14] |  -
7 | enable | TRUE | partitionA (project has parent) | parent partition [:14] | - 
8 | enable | TRUE | - (project has parent)| parent partition [:14] |  -
Updated testcases
  |   |   |   |   |  
9 | enable | TRUE | partitionA (project does not have parent) | existing project id [:14] | **Error
10 | enable | TRUE | - (project does not have parent) | existing project id [:14] | **Error

*Warning 
image

**Error
image
image

## Manual Testing
_Note :  `39acc1192457418ab2f5623e2e372344` is child project of `b572164adbcb489082dc4cfe0ad1d7e9`_ 
**_Below is test case 6 from above_**
Steps:
1) Set `a10-octavia.conf` with following configurations:
<pre>
[a10_global]
<b>use_parent_partition = False </b>

[hardware_thunder]
devices=[
{
 "project_id": "39acc1192457418ab2f5623e2e372344",   -------------------> child project 
 "ip_address": "10.0.0.XX",
 "device_name": "rack_vthunder",
 "username": "XX",
 "password": "XX",
 <b>"hierarchical_multitenancy": "enable"</b>
 }
]
</pre>
2) Restart the `a10-controller-worker.service`
3) Create LB in child project as follows : 
`openstack loadbalancer create --vip-subnet-id fd8effcf-f85c-498b-9bae-7ba36db4c01b --name SLB1 --project 39acc1192457418ab2f5623e2e372344`

4) Check on VThunder and openstack cli. It is created in child project partition.
image

image

** Above steps can be repeated for all the test cases mentioned above with different configurations

### DIFF
--- a/a10_octavia/controller/worker/tasks/a10_database_tasks.py
+++ b/a10_octavia/controller/worker/tasks/a10_database_tasks.py
@@ -195,6 +195,10 @@ class GetVThunderByLoadBalancer(BaseDatabaseTask):
                         vthunder.project_id)
                     if parent_project_id:
                         vthunder.partition_name = parent_project_id[:14]
+                    else:   ## TODO Should be handled via new db architechture
+                        LOG.warning("The parent project for project %s does not exist. "
+                                    "Configuration will be applied in project partition itself. ",
+                                    vthunder.project_id)
                 else:
                     LOG.warning("Hierarchical multitenancy is disabled, use_parent_partition "
                                 "configuration will not be applied for loadbalancer: %s",
--- a/a10_octavia/controller/worker/tasks/a10_database_tasks.py
+++ b/a10_octavia/controller/worker/tasks/a10_database_tasks.py
@@ -21,6 +21,7 @@ from oslo_config import cfg
 from oslo_log import log as logging
 from oslo_utils import uuidutils
 from taskflow import task
+from taskflow.types import failure
 
 from octavia.common import constants
 from octavia.db import api as db_apis
@@ -195,10 +196,10 @@ class GetVThunderByLoadBalancer(BaseDatabaseTask):
                         vthunder.project_id)
                     if parent_project_id:
                         vthunder.partition_name = parent_project_id[:14]
-                    else:   ## TODO Should be handled via new db architechture
-                        LOG.warning("The parent project for project %s does not exist. "
-                                    "Configuration will be applied in project partition itself. ",
-                                    vthunder.project_id)
+                    else:
+                        LOG.error("The parent project for project %s does not exist. ",
+                                  vthunder.project_id)
+                        raise exceptions.ParentProjectNotFound(vthunder.project_id)
                 else:
                     LOG.warning("Hierarchical multitenancy is disabled, use_parent_partition "
                                 "configuration will not be applied for loadbalancer: %s",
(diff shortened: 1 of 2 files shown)


a10_octavia/controller/worker/tasks/a10_database_tasks.py
```diff
@@ -189,15 +189,16 @@ def execute(self, loadbalancer):
...
-            elif use_parent_part and vthunder.hierarchical_multitenancy:
-                parent_project_id = utils.get_parent_project(
-                    vthunder.project_id)
-                if parent_project_id:
-                    vthunder.partition_name = parent_project_id[:14]
+            if use_parent_part:
+                if vthunder.hierarchical_multitenancy == 'enable':
+                    parent_project_id = utils.get_parent_project(
+                        vthunder.project_id)
+                    if parent_project_id:
```
hthompson-a10, line 196: We should raise a warning here or maybe even error out. The user has configured it to have the object created in the parent partition, but if none is found then we just create in the normal one. And then that normal partition can't be deleted without logging into the box....

hthompson-a10 (changes requested):

a10_octavia/controller/worker/tasks/a10_database_tasks.py
afrin-chakure-a10: Ok Will error it out here, for `None` `parent_project_id`

afrin-chakure-a10: Added `LOG.warning` and resources are created in existing project id partition itself. 
image

afrin-chakure-a10: **_Update_**:

Raised an Error here, instead of warning.

image

a10_octavia/controller/worker/tasks/a10_database_tasks.py
```diff
@@ -302,6 +308,19 @@ def execute(self, loadbalancer, vthunder_config):
                       loadbalancer.id)
             raise e
 
+    def revert(self, result, loadbalancer, vthunder_config, *args, **kwargs):
+        if isinstance(result, failure.Failure):
```
afrin-chakure-a10, line 312: This part seems to resolving the issue related to when revert happens from any other task and vthunder object is not. Maybe its not even needed. But for now keeping it for this PR.

OmkarTelee-A10 (approved):

hthompson-a10 (approved):

PR #183, opened 2020-09-18
[Stack 1492] Write memory in shared partition for Health monitor changes

OmkarTelee-A10:
## Issue Description
Whenever there is change in HM object, we must have to run "write mem" in shared partition as well.

## Jira Ticket
https://a10networks.atlassian.net/browse/STACK-1492

## Technical Approach
Enhanced writeMemory function to accept shared partition argument as an input. depending on which we can use it for multiple locations.

## Test Cases
- Test case from BUG only

## Manual Testing
- Checked how new code behaves when non-HM nouns 
- Checked write memory called with partition and in shared partition as well for HM nouns


hthompson-a10:
Please squash and rewrite the merge commits

hthompson-a10 (approved):

hthompson-a10 (changes requested):
Please squash and rewrite the merge commits

PR #186, opened 2020-09-18
[BUG][STACK-1619] Added update call if server already exists for related config updates

OmkarTelee-A10:
## Issue Descrption
When we are creating member in existing server with changes in config file, the server config (conn_limit, conn_resume) are not getting updated.

Severity Level : Critical

## Jira Ticket
https://a10networks.atlassian.net/browse/STACK-1619

## Technical Approach
- Added an update call if server already exists to make config changes in it
- Adjust try...catch block to capture ACOS_ERRORS like 'connectionError' for both the API calls

## Test Cases
1. Check config change for existing server
2. Check config change for new server
3. Check config change in between server created with same IP and different port
4. Check config change for new server with different IP
5. Check config change for SET function for pre-existing servers

## Manual Testing
Step1: Created a server with IP address 10.0.13.122 and port 81
!
slb server 67e06_10_0_13_122 10.0.13.122
port 81 tcp
!

Step2: Added following section in config and restarted the controller-worker
[server]
conn_limit=1100000
conn_resume=888877

Step3: Created another member with IP address 10.0.13.122 and port 82
slb server 67e06_10_0_13_122 10.0.13.122
  port 81 tcp
  port 82 tcp
  conn-limit 1100000
  conn-resume 888877
!

Step4: Added another server with different IP 
!
slb server 67e06_10_0_13_122 10.0.13.122
  port 81 tcp
  port 82 tcp
 conn-limit 1100000
  conn-resume 888877
!
slb server 67e06_10_0_13_123 10.0.13.123
  port 83 tcp
  port 84 tcp
 conn-limit 1100000
  conn-resume 888877
!

### DIFF
--- a/a10_octavia/controller/worker/tasks/server_tasks.py
+++ b/a10_octavia/controller/worker/tasks/server_tasks.py
@@ -50,12 +50,15 @@ class MemberCreate(task.Task):
             status = True
 
         try:
-            self.axapi_client.slb.server.create(server_name, member.ip_address, status=status,
-                                                server_templates=server_temp,
-                                                axapi_args=server_args)
-            LOG.debug("Successfully created member: %s", member.id)
-        except (acos_errors.Exists, acos_errors.AddressSpecifiedIsInUse):
-            pass
+            try:
+                self.axapi_client.slb.server.create(server_name, member.ip_address, status=status,
+                                                    server_templates=server_temp,
+                                                    axapi_args=server_args)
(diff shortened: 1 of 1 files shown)


rahulkatre-a10 (approved):

hthompson-a10 (approved):
I'm not a fan of nested try-blocks. I tried a number of ways to make it so we didn't need them, but since our API returns errors instead of empty payloads this is required.

PR #202, opened 2020-10-06
[STACK-1558]: VRID Floating IP For VIPs

rahulkatre-a10:
## Description
As a user, I want high availability to extend to my virtual servers so that I can ensure my load balancer stays up in the event of system shutdown/failure.
Initially, we were handling single VRID which is associated with a specific subnet. 
According to the new implementation, we can configure multiple VRIDs for LB resources residing in different subnets. And 
those VRIDs can be handled on LB resource object create, update,e and deletion actions.

## Jira Ticket
https://a10networks.atlassian.net/browse/STACK-1558

## Technical Approach
- Going through the existing VRID feature implementation, recognizing the module that can be reused for new requirements.
- Changes at acos-client for updating  VRRPA multiple floating IPs. 
- Made networking functions common for LB and members.
- For LB resource in question, deal with the list of VRIDs and the necessary action to be taken on the basis of action.
- For deletion LB resource object, before deletion check whether VRID is getting used for other LB resources added.
- Introduced multiple port removal network tasks for the pool, which internally uses Lb count and pool count for a specific subnet.
 
## Test Cases
- Added unit tests for multip-vrid delete of LB resource- network, database tasks.
- Added units tests for handling muti  VRID  for creating and update for LB resource- network and database task.

## Manual Testing
Load balancer resource   creation updation: 
- Create a load balancer in a specific subnet, all equivalent components(VRID floating IP, neutron port) should also get created. 
- Create LB in a different subnet, all equivalent components should get created.
- Create LB in the subnet where LB is already created, no VRID components should get created.

Load balancer resource deletion:
- Deleting LB with only that LB resource present in the subnet, deletes all VRID components.
- Deleting LB where another LB resides in the subnet, will not remove VRID components.
- Deleting LB where the member resides in the subnet, will not remove VRID components.

Pool deletion: 
- Pool with single member with specific subnet – Deletion of pool deleted member  and VRID and its associated components like port from neutron subnet. - PASSED 
- Pool with multiple member belongs to specific subnet -   Deletion of pool deleted all members and the attached VRID and its associated components like port from neutron subnet  - PASSED 
-  Pool with multiple member belongs to different subnet – Deletion of pool deleted all members and associated VRID and its components - PASSED 
- Pool with single member having subnet used by member belongs to other pool – Deletion of pool deleted all member belonging to pool being deleted, VRID and its components are not deleted. - PASSED 
-  Pool with multiple members having subnet used only in this pool for some members and subnet used in other for remaining members– Deletion of pool deleted all members but VRIDs for only those members whose subnet is used only for this pool members will be removed.  - PASSED


hthompson-a10:
This PR needs to be pointed at vrid feature branch. Then we can merge that branch into master.

PR #209, opened 2020-10-15
[STACK-1690]: Wrapping up loadbalancer description into quotes.

NehaKembalkarA10:
## Description
- Adding quotes around the loadbalancer description sothat there will be sync between OpenStack side and vThunder side for any description with having leading or trailing spaces to it.

## Jira Ticket
https://a10networks.atlassian.net/browse/STACK-1690

## Technical Approach
- If there is no description parameter provided while creating a loadbalancer, then None will be set as the description and there will not be any description associated with the loadbalancer on both sides.
- If description provided while creating a loadbalancer is "      " or "", then there will not be any description associated with the loadbalancer on both sides.
- Otherwise any other description will get associated exactly same as that of provided while loadbalancer creation, wrapped in quotes(""), on both sides. 

## Manual Testing
Testcases required for this story:

image

image

### DIFF
--- a/a10_octavia/controller/worker/tasks/virtual_server_tasks.py
+++ b/a10_octavia/controller/worker/tasks/virtual_server_tasks.py
@@ -37,7 +37,12 @@ class LoadBalancerParent(object):
         arp_disable = CONF.slb.arp_disable
         vrid = CONF.slb.default_virtual_server_vrid
         desc = loadbalancer.description
-        desc = "" if str(desc).isspace() else desc
+        if not desc:
+            desc = None
+        elif str(desc).isspace() or (not str(desc)):
+            desc = ""
+        else:
+            desc = '"{}"'.format(desc)
 
         set_method(
             loadbalancer.id,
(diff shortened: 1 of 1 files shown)


a10_octavia/controller/worker/tasks/virtual_server_tasks.py
```diff
@@ -37,7 +37,12 @@ def set(self, set_method, loadbalancer):
         arp_disable = CONF.slb.arp_disable
         vrid = CONF.slb.default_virtual_server_vrid
         desc = loadbalancer.description
-        desc = "" if str(desc).isspace() else desc
+        if not desc:
+            desc = None
+        elif str(desc).isspace() or not str(desc):
```
rahulkatre-a10, line 42: Is there any case for which `not desc`  and `not str(desc)`  results different value.

NehaKembalkarA10: Yes...
When we don't provide description parameter while creating lb, then `not desc` = **True**  and `not str(desc)` = **False**
Because at that time we get `desc` as empty but `str(desc)` as 'None'

hthompson-a10 (approved):

afrin-chakure-a10 (approved):

rahulkatre-a10 (approved):

PR #218, opened 2020-10-23
[BUG][STACK-1691]: [vrid_floating_ip] different subnet vrid_floating_ip is deleted wrongly

rahulkatre-a10:
## Description
- Severity Level: High
- Required: Issue: The VRID feature is not built by keeping HMT and use parent partition in mind. Previously the consideration was user mapped with project and project is mapped to partition so the VRID list from DB, LB count for the subnet and, member count subnet are fetched based on a project which caused issues when the partition is shared among projects. 
 Now with this PR, we use to partition and the projects using that partition for fetching VRID list, LB count, and member count. 

## Jira Ticket
https://a10networks.atlassian.net/browse/STACK-1691

## Technical Approach
- Recognized the other cases of failure like member delete, LB delete and pool delete.
- With previous PR we fixed the issue fo LB and member creation. 
- Through bug STACK-1691, we recognized the fetch VRID list, LB count, member count needed to be integrated with delete flow.
- Added new DB task  "GetProjectsForPartition" for fetching projects using a partition in question.
- Updated existing tasks for fetching the VRID list, LB count, and member count to accept the project list and accordingly update query to DB.

## Test Cases
- Added test for testing "GetProjectsForPartition", fetching project list for a specific partition.

## Manual Testing
- Created LB in 2 different child projects with below config:
     hierarchical_multitenancy="enable"
     use_parent_partition=True
 Load balancers created:
   "lc1" -> "cadmin"(project) 
   "lcc1"-> "ccadmin"(project)

Testing deletion of LB:
- Deleted lc1, deleted the equivalent virtual server from vThunder and VRID FIP also getting removed, neutron port is getting removed
- Deleted lcc1, deleted equivalent virtual server,  VRID FIP removed, and neutron port is also getting removed.

### DIFF
--- a/a10_octavia/controller/worker/tasks/a10_database_tasks.py
+++ b/a10_octavia/controller/worker/tasks/a10_database_tasks.py
@@ -383,24 +383,14 @@ class CreateSpareVThunderEntry(BaseDatabaseTask):
 
 class GetVRIDForLoadbalancerResource(BaseDatabaseTask):
 
-    def execute(self, lb_resource):
+    def execute(self, partition_project_list):
         try:
-            project_id = lb_resource.project_id
-            vthunder_conf = CONF.hardware_thunder.devices[project_id]
-            if vthunder_conf.hierarchical_multitenancy == 'enable' and CONF.a10_global.use_parent_partition:
-                partition_name = self.vthunder_repo.get_partition_for_project(
-                    db_apis.get_session(), project_id=project_id)
-                project_ids = self.vthunder_repo.get_project_list_using_partition(
-                    db_apis.get_session(), partition_name=partition_name)
-
-            else:
-                project_ids = [project_id]
             vrid_list = self.vrid_repo.get_vrid_from_project_ids(
-                db_apis.get_session(), project_ids=project_ids)
+                db_apis.get_session(), project_ids=partition_project_list)
             return vrid_list
         except Exception as e:
-            LOG.exception("Failed to get VRID list for given project  %s due to %s",
-                          lb_resource.project_id, str(e))
+            LOG.exception("Failed to get VRID list for given project list  %s due to %s",
+                          partition_project_list, str(e))
             raise e
 
 
@@ -451,21 +441,11 @@ class UpdateVRIDForLoadbalancerResource(BaseDatabaseTask):
 
 
 class CountLoadbalancersInProjectBySubnet(BaseDatabaseTask):
-    def execute(self, lb_resource, subnet):
+    def execute(self, subnet, partition_project_list):
         try:
-            project_id = lb_resource.project_id
-            vthunder_conf = CONF.hardware_thunder.devices[project_id]
-            if vthunder_conf.hierarchical_multitenancy == 'enable' and CONF.a10_global.use_parent_partition:
-                partition_name = self.vthunder_repo.get_partition_for_project(
-                    db_apis.get_session(), project_id=project_id)
-                project_ids = self.vthunder_repo.get_project_list_using_partition(
-                    db_apis.get_session(), partition_name=partition_name)
-            else:
-                project_ids = [project_id]
-
             return self.loadbalancer_repo.get_lb_count_by_subnet(
                 db_apis.get_session(),
-                project_ids=project_ids, subnet_id=subnet.id)
+                project_ids=partition_project_list, subnet_id=subnet.id)
         except Exception as e:
             LOG.exception("Failed to get LB count for subnet %s due to %s ",
                           subnet.id, str(e))
@@ -473,20 +453,11 @@ class CountLoadbalancersInProjectBySubnet(BaseDatabaseTask):
 
 
 class CountMembersInProjectBySubnet(BaseDatabaseTask):
-    def execute(self, lb_resource, subnet):
+    def execute(self, subnet, partition_project_list):
         try:
-            project_id = lb_resource.project_id
-            vthunder_conf = CONF.hardware_thunder.devices[project_id]
-            if vthunder_conf.hierarchical_multitenancy == 'enable' and CONF.a10_global.use_parent_partition:
-                partition_name = self.vthunder_repo.get_partition_for_project(
-                    db_apis.get_session(), project_id=project_id)
-                project_ids = self.vthunder_repo.get_project_list_using_partition(
-                    db_apis.get_session(), partition_name=partition_name)
-            else:
-                project_ids = [project_id]
             return self.member_repo.get_member_count_by_subnet(
                 db_apis.get_session(),
-                project_ids=project_ids, subnet_id=subnet.id)
+                project_ids=partition_project_list, subnet_id=subnet.id)
         except Exception as e:
             LOG.exception(
                 "Failed to get LB member count for subnet %s due to %s",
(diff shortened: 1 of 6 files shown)


a10_octavia/controller/worker/tasks/a10_database_tasks.py
```diff
@@ -690,13 +639,42 @@ def execute(self, member_list):
...
+            project_id = lb_resource.project_id
+            partition_name = self.vthunder_repo.get_partition_for_project(
+                db_apis.get_session(), project_id=project_id)
+            if not partition_name:
+                # There is a case when error occured while creating LB causes vthunder
+                # entry to be removed, hence need to fetch partition name using parent project id
+                parent_project_id = utils.get_parent_project(project_id)
+                if parent_project_id:
+                    partition_name = parent_project_id[:14]
+                l2_project.append(project_id)
```
NehaKembalkarA10, line 671: As the `partition_project_list` can be got using `get_project_list_using_partition` method which requires only the current partition_name to be known, for getting all project ids under that partition, why again this `l2_project[]` is required?

rahulkatre-a10: I observed a case when vthunder entry getting deleted via revert when the creation of lb fails. In that case, I explicitly insert the project in operation in the project's list.

rahulkatre-a10: Added changes.

NehaKembalkarA10 (dismissed):

a10_octavia/controller/worker/tasks/a10_database_tasks.py
```diff
@@ -313,23 +312,6 @@ def execute(self, loadbalancer, vthunder_config):
                 loadbalancer.id)
             raise e
 
-    def revert(self, result, loadbalancer, vthunder_config, *args, **kwargs):
```
hthompson-a10, line 316: Why was this revert removed?

a10_octavia/controller/worker/tasks/a10_database_tasks.py
```diff
@@ -690,13 +639,29 @@ def execute(self, member_list):
...
                     if pool_count_subnet <= 1 and lb_count_subnet == 0:
                         subnet_list.append(member.subnet_id)
                     member_subnet.append(member.subnet_id)
             return subnet_list
         except Exception as e:
             LOG.exception("Failed to get subnet list for members: %s", str(e))
             raise e
+
+
+class GetProjectsForPartition(BaseDatabaseTask):
```
hthompson-a10, line 654: The naming on this needs to better convey that it's meant for child projects

ie `GetChildProjectsOfParentPartition`

a10_octavia/db/repositories.py
```diff
@@ -303,12 +303,17 @@ def get_all_deleted_expiring(self, session, exp_age):
         return id_list
 
     def get_partition_for_project(self, session, project_id):
-        return session.query(self.model_class).filter(
-            self.model_class.project_id == project_id).first().partition_name
+        vthunder_project = self.get_vthunder_by_project_id(session, project_id)
+        if vthunder_project:
```
hthompson-a10, line 307: This should never be an issue. We should never store a Thunder entry unless it has a partition name already

a10_octavia/tests/unit/controller/worker/tasks/test_a10_database_tasks.py
```diff
@@ -356,5 +355,15 @@ def test_get_subnet_for_deletion_with_multiple_pool_lb(self):
         mock_subnets.member_repo.get_pool_count_subnet.return_value = 2
         mock_subnets.loadbalancer_repo.get_lb_count_by_subnet = mock.Mock()
         mock_subnets.loadbalancer_repo.get_lb_count_by_subnet.return_value = 2
-        subnet_list = mock_subnets.execute([MEMBER_1, MEMBER_2])
+        subnet_list = mock_subnets.execute(
+            [MEMBER_1, MEMBER_2], [a10constants.MOCK_PROJECT_ID])
         self.assertEqual([], subnet_list)
+
+    def test_get_projects_for_partition(self):
```
hthompson-a10, line 362: This only handle the success case. We need the failure cases to be tested as well. Have one of the calls have the side effect of an error and ensure it is caught.

hthompson-a10 (changes requested):

a10_octavia/controller/worker/tasks/a10_database_tasks.py
rahulkatre-a10: While testing I was seeing an error where when creating LB if LB goes into an error state. And when we try to delete an LB is was giving an error. This was due to this revert statement which causes vThunder entry to be removed from DB. and Delete operation does require vthunder object to make axapi calls. 
@afrin-chakure-a10 , Please confirm

a10_octavia/controller/worker/tasks/a10_database_tasks.py
rahulkatre-a10: Here we are not checking multitenancy settings, hence it will be used for all the projects no matter whether it is a parent or children.

a10_octavia/db/repositories.py
rahulkatre-a10: This was the issue occurring due to the revert statement which I removed "CreateRackVthunderEntry". Since I removed that this condition is not required here.
will remove it.

a10_octavia/controller/worker/tasks/a10_database_tasks.py
hthompson-a10: It's about intention though. We never want multiple projects to be tied to one partition. The only time this is okay is if these are child projects.

So we intend this class to be used in that way and should name it as such

rahulkatre-a10: 0k will change the function name.

a10_octavia/tests/unit/controller/worker/tasks/test_a10_database_tasks.py
rahulkatre-a10: The task only has calls to DB. It does not have any other flow except DB queries.

a10_octavia/controller/worker/tasks/a10_database_tasks.py
hthompson-a10: Instead of removing the revert here, let's create another bug to fix the delete operation issue

https://github.com/a10networks/a10-octavia/blob/master/a10_octavia/controller/worker/tasks/virtual_server_tasks.py#L87

We already check to see if there is a Thunder present for the virtual server delete call. So it is probably the VRRP task or VE IP task causing the issue.

hthompson-a10:
Please squash this commit "Incorporated review comments" into a previous commit as it doesn't have any info in the message

a10_octavia/controller/worker/tasks/a10_database_tasks.py
rahulkatre-a10: https://github.com/a10networks/a10-octavia/blob/d14308234f4956abda0a3177059d176143d73666/a10_octavia/controller/worker/tasks/a10_database_tasks.py#L316, This is a similar task here as well we don't have revert, previously revert was not there It was recently added on around 11 September 2020.

hthompson-a10: I don't agree with that recent change and it was my bad for not catching it. We don't want to run the risk of a orphaned Thunder entry. It would be like having an orphaned Amphora.

https://github.com/openstack/octavia/blob/master/octavia/controller/worker/v2/tasks/database_tasks.py#L89

You can see upstream where they have a revert in this scenario for the Amphora. In our case, the Amphora == Thunder

rahulkatre-a10: Have added the revert.

hthompson-a10 (approved):

afrin-chakure-a10 (approved):

PR #223, opened 2020-11-02
[STACK-1654] healthmonitor support POST method

ytsai-a10:
## Description
Serverity Level: Medium
Description:
Failed to create health monitor on Thunder when HTTP method is POST. 
Because the axapi request failed. The axapi expect following for POST healthy monitor:
- use post-path instead of url-path
- require http-postdata and post-type key in json

example:
```
{"monitor": { "name":"1005", "retry":3, "up-retry":1, "override-port":80, "interval":4, "timeout":3, "method": { "http": {"http":1, "http-port":80, "http-expect":1, "http-response-code":"200", "http-url":1, "url-type":"POST", "post-path":"/", "post-type":"postdata", "http-postdata":"abc=1" }}}}
```

## Jira Ticket
https://a10networks.atlassian.net/browse/STACK-1654

## Technical Approach
- in /etc/a10/a10-octavia.conf, add new option **post_data** for **a10_health_manager**
- In acos-client we will use the value of **post_data** for **http-postdata** key in json.
- also change **url-path** to **post-path** for POST case
- add "post-type": "postdata" in json

see also: https://github.com/a10networks/acos-client/pull/307/files

## Config Changes

<pre>
<b>[health_monitor]
post_data = "abc=1"
</b>
</pre>

## Test Cases
- create healthmonitor with http POST method
- change options in healthmonitor 
- delete healthmonitor
- create healthmonitor with http GET method
- change options in healthmonitor 
- delete healthmonitor

## Manual Testing

```
openstack loadbalancer healthmonitor create --delay 3 --timeout 3 --max-retries 2 --http-method POST --type HTTP  sg1 --name hm1
openstack loadbalancer healthmonitor set --max-retries 5 hm1
openstack loadbalancer healthmonitor delete hm1
openstack loadbalancer healthmonitor create --delay 3 --timeout 3 --max-retries 2 --type HTTP  sg1 --name hm2
openstack loadbalancer healthmonitor set --max-retries 5 hm2
openstack loadbalancer healthmonitor delete hm2
```


hthompson-a10:
@ytsai-a10 Looks like a minor change is required in unit tests. The `post_data=None` needs to be added to the expected call

```
Expected: create('mock-hm-1', <Mock name='mock.slb.hm.TCP' id='139869620411408'>, 7, 3, 8, axapi_args={}, expect_code=None, method=None, port=<ANY>, url=None)

Actual: create('mock-hm-1', <Mock name='mock.slb.hm.TCP' id='139869620411408'>, 7, 3, 8, axapi_args={}, expect_code=None, method=None, port=<ANY>, post_data=None, url=None)
```

a10_octavia/controller/worker/tasks/health_monitor_tasks.py
```diff
@@ -44,13 +44,15 @@ def execute(self, listeners, health_mon, vthunder):
         args = utils.meta(health_mon, 'hm', {})
 
         try:
+            pdata = CONF.a10_health_manager.post_data
```
hthompson-a10, line 47: Bit of a nitpick, but it is cleaner here to just define it as `post_data` so long as it doesn't push us over line length limit.

We don't have to use the variable anywhere beyond this one section so there isn't a need to truncate

hthompson-a10 (changes requested):

a10_octavia/controller/worker/tasks/health_monitor_tasks.py
```diff
@@ -122,11 +124,12 @@ def execute(self, listeners, health_mon, vthunder, update_dict):
             expect_code = health_mon.expected_codes
         args = utils.meta(health_mon, 'hm', {})
         try:
+            post_data = CONF.a10_health_manager.post_data
```
afrin-chakure-a10, line 127: Is it alright to take this `post_data` configured from  `[a10_health_manager]` section? Because according to me, `health monitor` created via openstack cli is part of `a10-controller-worker.service` service whereas, things configured under `[a10_health_manager]` section is for separate service called `a10-health_manager.service`

I would suggest we should read it from another section called `[health_monitor]` similar to  `[listener]`, `[slb]` etc. And from there we would read `post_data`

afrin-chakure-a10:
```
[health_monitor]
post_data = "abc=1"
```

ytsai-a10: changed as the comment

afrin-chakure-a10 (dismissed):
looks good 👍

afrin-chakure-a10 (approved):

NehaKembalkarA10 (approved):

a10_octavia/controller/worker/tasks/health_monitor_tasks.py
ytsai-a10: updated

hthompson-a10 (approved):

PR #274, opened 2020-12-28
STACK-1900: Creation of nat-pools in nat-pool-list flavor

NehaKembalkarA10:
## Description
For given nat-pool-list flavor, NAT pool will be created on Thunder when loadbalancer is created.

## Jira Ticket
https://a10networks.atlassian.net/browse/STACK-1900

## Technical Approach
- When nat-pool-list flavor or both nat-pool and nat-pool-list is passed while creating a loadbalancer:
        - if the specified nat-pool with the same range does not exists on thunder, the nat-pool given in the flavor gets created.
        - if the specified nat-pool with the same range exists on thunder, the nat-pool will not get created on thunder as it already
          exists.
        - if the specified nat-pool with the different range exists on thunder, an error will be raised with a message as "The nat-pool
          already exists on thunder".

## Test Cases
- Given a nat-pool-list flavor, when a loadbalancer is created, the nat-pool will get created on thunder if the same nat-pool does not exist on it.
- Given a nat-pool-list flavor, when a loadbalancer is created, the nat-pool will not get created on thunder if the same nat-pool with same range exists on thunder.
- Given a nat-pool-list flavor, when a loadbalancer is created, an error will be raised if the same nat-pool with different range exists on thunder.

## Manual Testing
Flavor used for testing:
1.

```
{ 
    "nat-pool":{ 
        "pool-name":"pool1", 
        "start-address":"172.16.1.101", 
        "end-address":"172.16.1.104", 
        "netmask":"/24", 
        "gateway":"172.16.1.1" 
    }, 
    "nat-pool-list":[ 
        { 
            "pool-name":"pool2", 
            "start-address":"172.16.2.101", 
            "end-address":"172.16.2.104", 
            "netmask":"/24", 
            "gateway":"172.16.2.1" 
        }, 
        { 
            "pool-name":"pool3", 
            "start-address":"172.16.3.101", 
            "end-address":"172.16.3.104", 
            "netmask":"/24", 
            "gateway":"172.16.3.1" 
        } 
    ],
}
```
```
openstack loadbalancer flavorprofile create --name fp_snat_list --provider a10 --flavor-data '{"nat-pool":{"pool-name":"pool1", "start-address":"172.16.1.101", "end-address":"172.16.1.102", "netmask":"/24", "gateway":"172.16.1.1"}, "nat-pool-list":[{"pool-name":"pool2", "start-address":"172.16.2.101", "end-address":"172.16.2.102", "netmask":"/24", "gateway":"172.16.2.1"}, {"pool-name":"pool3", "start-address":"172.16.3.101", "end-address":"172.16.3.102", "netmask":"/24", "gateway":"172.16.3.1"}]}'

openstack loadbalancer flavor create --name f_snat_list --flavorprofile fp_snat_list --description "flaovr all test1" --enable

openstack loadbalancer create --flavor f_snat_list --vip-subnet-id 11dcb328-f947-4b5f-a789-5ae46118bb81 --name lb1

```
Result:
image

2.
```
{
    "nat-pool-list":[ 
        { 
            "pool-name":"pool1", 
            "start-address":"172.16.2.101", 
            "end-address":"172.16.2.104", 
            "netmask":"/24", 
            "gateway":"172.16.2.1" 
        }, 
        { 
            "pool-name":"pool2", 
            "start-address":"172.16.3.101", 
            "end-address":"172.16.3.104", 
            "netmask":"/24", 
            "gateway":"172.16.3.1" 
        } 
    ],
}
```

```
openstack loadbalancer flavorprofile create --name fp_snat_list2 --provider a10 --flavor-data '{"nat-pool-list":[{"pool-name":"pooln1", "start-address":"172.16.2.101", "end-address":"172.16.2.102", "netmask":"/24", "gateway":"172.16.2.1"}, {"pool-name":"pooln2", "start-address":"172.16.3.101", "end-address":"172.16.3.102", "netmask":"/24", "gateway":"172.16.3.1"}]}'

openstack loadbalancer flavor create --name f_snat_list2 --flavorprofile fp_snat_list2 --description "flaovr all test2" --enable

openstack loadbalancer create --flavor f_snat_list2 --vip-subnet-id 11dcb328-f947-4b5f-a789-5ae46118bb81 --name lb2

```
Result:
image

### DIFF
--- a/a10_octavia/controller/worker/tasks/nat_pool_tasks.py
+++ b/a10_octavia/controller/worker/tasks/nat_pool_tasks.py
@@ -50,8 +50,9 @@ class NatPoolCreate(task.Task):
                             self.axapi_client.nat.pool.create(
                                 pool_name, start_address, end_address, netmask, gateway=gateway)
                         except(acos_errors.Exists) as e:
-                            LOG.exception(
-                                "The specified nat-pool already exists on thunder")
+                            LOG.exception("Nat-pool with name %s already exists on partition %s of "
+                                          "thunder device %s",
+                                          pool_name, vthunder.partition_name, vthunder.ip_address)
                             raise e
                     else:
                         continue
@@ -73,5 +74,7 @@ class NatPoolCreate(task.Task):
                     self.axapi_client.nat.pool.create(
                         pool_name, start_address, end_address, netmask, gateway=gateway)
                 except(acos_errors.Exists) as e:
-                    LOG.exception("The specified nat-pool already exists on thunder")
+                    LOG.exception("Nat-pool with name %s already exists on partition %s of "
+                                  "thunder device %s",
+                                  pool_name, vthunder.partition_name, vthunder.ip_address)
                     raise e
--- a/a10_octavia/controller/worker/tasks/nat_pool_tasks.py
+++ b/a10_octavia/controller/worker/tasks/nat_pool_tasks.py
@@ -29,24 +29,49 @@ class NatPoolCreate(task.Task):
     def execute(self, loadbalancer, vthunder, flavor_data=None):
         device_pool = None
         if flavor_data:
+            natpool_flavor_list = flavor_data.get('nat_pool_list')
             natpool_flavor = flavor_data.get('nat_pool')
+            if natpool_flavor_list:
+                for i in range(len(natpool_flavor_list)):
+                    if self.axapi_client.nat.pool.exists(natpool_flavor_list[i]['pool_name']):
+                        device_pool = self.axapi_client.nat.pool.get(
+                            natpool_flavor_list[i]['pool_name'])
+                    if not (device_pool and
+                            (device_pool['pool']['start-address'] ==
+                                natpool_flavor_list[i]['start_address'] and
+                                device_pool['pool']['end-address'] ==
+                                natpool_flavor_list[i]['end_address'])):
+                        pool_name = natpool_flavor_list[i]['pool_name']
+                        start_address = natpool_flavor_list[i]['start_address']
+                        end_address = natpool_flavor_list[i]['end_address']
+                        netmask = natpool_flavor_list[i]['netmask']
+                        gateway = natpool_flavor_list[i]['gateway']
+                        try:
+                            self.axapi_client.nat.pool.create(
+                                pool_name, start_address, end_address, netmask, gateway=gateway)
+                        except(acos_errors.Exists) as e:
+                            LOG.exception(
+                                "The specified nat-pool already exists on thunder")
+                            raise e
+                    else:
+                        continue
             if natpool_flavor:
                 if self.axapi_client.nat.pool.exists(natpool_flavor['pool_name']):
-                    device_pool = self.axapi_client.nat.pool.get(natpool_flavor['pool_name'])
+                    device_pool = self.axapi_client.nat.pool.get(
+                        natpool_flavor['pool_name'])
                 if (device_pool and
-                        (device_pool['pool']['start-address'] == natpool_flavor['start_address']
-                            and device_pool['pool']['end-address'] == natpool_flavor['end_address']
-                         )):
+                    (device_pool['pool']['start-address'] == natpool_flavor['start_address']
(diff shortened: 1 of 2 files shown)


ytsai-a10 (approved):

a10_octavia/controller/worker/tasks/nat_pool_tasks.py
```diff
@@ -29,24 +29,49 @@ class NatPoolCreate(task.Task):
...
+                        start_address = natpool_flavor_list[i]['start_address']
+                        end_address = natpool_flavor_list[i]['end_address']
+                        netmask = natpool_flavor_list[i]['netmask']
+                        gateway = natpool_flavor_list[i]['gateway']
+                        try:
+                            self.axapi_client.nat.pool.create(
+                                pool_name, start_address, end_address, netmask, gateway=gateway)
+                        except(acos_errors.Exists) as e:
+                            LOG.exception(
+                                "The specified nat-pool already exists on thunder")
```
hthompson-a10, line 54: Though this is more of a defensive code section (given that we should never trigger this exception), it should still include the nat-pool name, flavor-id, lb-id. Ex:

"Nat-pool with name {} already exists on partition {} of thunder device {}"

Should this worst case occur, we still want an operator to know how to recover on their end.

hthompson-a10 (changes requested):
Another edge case we have here is the following:

1) User defines a flavorprofile that has natpool with poolA
2) User updates the flavorprofile with a natpool list that includes poolA

This would end up causing a conflict that the code is currently structured to handle, but that we need to have test for.

NehaKembalkarA10:
> Another edge case we have here is the following:
> 
> 1. User defines a flavorprofile that has natpool with poolA
> 2. User updates the flavorprofile with a natpool list that includes poolA
> 
> This would end up causing a conflict that the code is currently structured to handle, but that we need to have test for.

NehaKembalkarA10:
> Another edge case we have here is the following:
> 
> 1. User defines a flavorprofile that has natpool with poolA
> 2. User updates the flavorprofile with a natpool list that includes poolA
> 
> This would end up causing a conflict that the code is currently structured to handle, but that we need to have test for.

NehaKembalkarA10:
> Another edge case we have here is the following:
> 
> 1. User defines a flavorprofile that has natpool with poolA
> 2. User updates the flavorprofile with a natpool list that includes poolA
> 
> This would end up causing a conflict that the code is currently structured to handle, but that we need to have test for.

@hthompson-a10 , the below condition is testing this case if poolA of natpool flavor already exists on thunder.
`if self.axapi_client.nat.pool.exists(natpool_flavor_list[i]['pool_name'])`
After this test, there is again 1 condition next to this is checking if the updated values of poolA  contain same start-address and end-address or not and executing further things depending on it.

If the your mentioned 2 points are happening at a same time 1st time with poolA not existing on thunder, then
natpool poolA with configurations provided in nat-pool-list flavorprofile will be created first and then the below condition check will happen for natpool flavorprofile
`if self.axapi_client.nat.pool.exists(natpool_flavor['pool_name']):`
After this test, there is again 1 condition next to this is checking if the updated values of poolA  contain same start-address and end-address or not and executing further things depending on it.

a10_octavia/controller/worker/tasks/nat_pool_tasks.py
NehaKembalkarA10: Changed the exception message.

ytsai-a10: @hthompson-a10 
openstack didn't allow update (or set) flavorprofile once it is in use. So, this is fine.

`stack@ytsai-openstack:~/source/a10-octavia/snat$ openstack loadbalancer flavorprofile set --flavor-data '' fp_snat
Flavor profile 93bec7f8-9f39-44ee-bc75-1d0e8f8fe8f6 is in use and cannot be modified. (HTTP 409) (Request-ID: req-c5eb45d8-6bfe-4aeb-b869-a8d847ee4c92)
`

hthompson-a10:
@NehaKembalkarA10  I meant a unit test

PR #310, opened 2021-01-25
[BUG][STACK-2046]: Corrected lb_method configuring for lb-algorithm SOURCE_IP for pool

NehaKembalkarA10:
## Description
- Severity Level: Medium
- For pool, lb-algorithm SOURCE_IP was getting configured as src-ip-only-hash instead of src-ip-hash on thunder.

## Jira Ticket
https://a10networks.atlassian.net/browse/STACK-2046

## Technical Approach
Modified the lb_method for `SOURCE_IP` from `SOURCE_IP_HASH_ONLY` to `SOURCE_IP_HASH` to get method **src-ip-hash** on thunder.

## Test Cases
Given lb-algorithm as SOURCE_IP, when a pool is created with it, then method src-ip-hash will get configured to it on thunder

## Manual Testing
1. Create a loadbalancer.
openstack loadbalancer create --name lb1 --vip-subnet-id provider-vlan-11-subnet

2. Create a listener.
openstack loadbalancer listener create --name l1 --protocol HTTP --protocol-port 82 lb1

3. Create a pool with lb-algorithm as SOURCE_IP
openstack loadbalancer pool create --protocol HTTP --lb-algorithm SOURCE_IP --listener l1 --name pool1

Result:
image

### DIFF
--- a/a10_octavia/common/openstack_mappings.py
+++ b/a10_octavia/common/openstack_mappings.py
@@ -30,7 +30,7 @@ def service_group_lb_method(c, os_method):
     lb_methods = {
         'ROUND_ROBIN': z.ROUND_ROBIN,
         'LEAST_CONNECTIONS': z.LEAST_CONNECTION,
-        'SOURCE_IP': z.SOURCE_IP_HASH_ONLY,
+        'SOURCE_IP': z.SOURCE_IP_HASH,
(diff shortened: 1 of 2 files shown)


ytsai-a10 (commented):
according to the help:
```
  src-ip-hash                            Load-balancing based on only Src IP and Port hash
  src-ip-only-hash                       Load-balancing based on only Src IP hash
```
The current behavior is fine, we don't need to change it.

PR #334, opened 2021-03-03
STACK[2015][Dynamic Interface Attachment]: Dynamic Interface Attachment Detection for loadbalancers

NehaKembalkarA10:
## Description
Added support for dynamic interface attachment detection in case of loadbalancer creation.

## Jira Ticket
https://a10networks.atlassian.net/browse/STACK-2155
https://a10networks.atlassian.net/browse/STACK-2159

## Related PR:
https://github.com/a10networks/acos-client/pull/337

## Technical Approach
- Modified `AmphoraePostVIPPlug` task to perform reload on second onwards loadbalancer creation on a vThunder having acos_version greater than or equal to 5.2.0.
- Modified `AmphoraePostVIPPlug` task to perform reboot on second onwards loadbalancer creation on a vThunder having acos_version less than 5.2.0.
- Added a column **acos_version** into vthunders table to save acos_version of vthunder.
- Added a task **UpdateAcosVersionInVthunderEntry** for saving acos_version into a column acos_version of vthunders table and called the task into `a10_load_balancer_flows` before `AmphoraePostVIPPlug` for getting acos_version of the vthunder.
- Added logic in it to call an axapi for getting and saving acos version in vthunder entry on 1st loadbalancer creation for a particular project or otherwise get the acos version for other loadbalancers with the same project id from the existing vthunder entry only, sothat there is no need to call the api each time.
- Modified `EnableInterface` task to enable any attached interface on particular loadbalancer creation.
- Added tasks `ReloadLoadBalancer`, `AllocateVIP` and `UpdateVIPAfterAllocation` to `vthunder_flows` to support dynamic interface attachment 2nd loadbalancer onwards.
- Added logic to `a10_load_balancer_flows` to escape reboot/reload for 1st loadbalancer creation as well as for subsequent loadbalancers having same subnet-id as that of existing one, as reboot/reload is not required in these cases.
- To handle this case, added a method **get_lb_exists_flag** which returns True if there is no any existing loadbalancer with same subnet and project id of current is present on the vthunder.

## Test Cases
- Given a loadbalancer, when acos_version of its vthunder is greater than or equal to 5.2, then reload should be performed during loadbalancer creation.
- Given a loadbalancer, when acos_version of its vthunder is less than 5.2, then reboot should be performed during loadbalancer creation. 
- Given a loadbalancer, when acos_version of its vthunder is greater than or equal to 5.2, then an axapi call should be performed to get acos_version for updating it in db.
-  Given a loadbalancer, when acos_version of its vthunder is less than 5.2, then an axapi call should not be performed to get acos_version for updating it in db.

## Manual Testing
**Case 1: Creation of loadbalancers using admin project.**

A vThunder instance will get booted for below loadbalancers 
1. Create lb1 with subnet as vip_subnet1 
        - Reload/Reboot will not get performed 
        - AXAPI call will be executed for saving acos_version in db 
        - Interface ethernet1 will get attached and enabled on the vThunder
2. Create lb2 with subnet as vip_subnet1 
        - Reload/Reboot will not get performed
        - No AXAPI call for saving acos_version in db, Instead acos_version will be fetched from existing vthunder entry created for 
           lb1 and saved in db.
        - No new interface will get attached for lb2
3. Create lb3 with subnet as vip_subnet1 
        - Reload/Reboot will not get performed
        - No AXAPI call for saving acos_version in db, Instead acos_version will be fetched from existing vthunder entry created for 
           lb1 and saved in db.
        - No new interface will get attached for lb3
4. Create lb4 with subnet as vip_subnet2
        - Reload/Reboot will get performed
        - No AXAPI call for saving acos_version in db, Instead acos_version will be fetched from existing vthunder entry created for 
           lb1 and saved in db.
        - New interface ethernet2 will get attached and enabled for lb4 on vThunder.
  
image

**Case 2: Creation of loadbalancers using demo project.**

A new vThunder instance will get booted for below loadbalancers 
1. Create lb1 with subnet as vip_subnet1 
        - Reload/Reboot will not get performed 
        - AXAPI call will be executed for saving acos_version in db 
        - Interface ethernet1 will get attached and enabled on the vThunder
2. Create lb2 with subnet as vip_subnet1 
        - Reload/Reboot will not get performed
        - No AXAPI call for saving acos_version in db, Instead acos_version will be fetched from existing vthunder entry created for 
           lb1 and saved in db.
        - No new interface will get attached for lb2
3. Create lb3 with subnet as vip_subnet1 
        - Reload/Reboot will not get performed
        - No AXAPI call for saving acos_version in db, Instead acos_version will be fetched from existing vthunder entry created for 
           lb1 and saved in db.
        - No new interface will get attached for lb3
4. Create lb4 with subnet as vip_subnet2
        - Reload/Reboot will get performed
        - No AXAPI call for saving acos_version in db, Instead acos_version will be fetched from existing vthunder entry created for 
           lb1 and saved in db.
        - New interface ethernet2 will get attached and enabled for lb4 on vThunder.

image


a10_octavia/controller/worker/tasks/vthunder_tasks.py
```diff
@@ -103,11 +103,19 @@ class AmphoraePostVIPPlug(VThunderBaseTask):
     def execute(self, loadbalancer, vthunder):
         """Execute get_info routine for a vThunder until it responds."""
         try:
-            self.axapi_client.system.action.write_memory()
-            self.axapi_client.system.action.reboot()
-            LOG.debug("Waiting for 30 seconds to trigger vThunder reboot.")
-            time.sleep(30)
-            LOG.debug("Successfully rebooted vThunder: %s", vthunder.id)
+            if "5.2" in vthunder.acos_version[:3]:
```
ytsai-a10, line 106: For ACOS version, it may change to 5.3.0, 5.4.0 or 6.0.0...etc. 
Can we find a way to compare version > 5.2 and write an API to do this? 

Here is my thought:
- Provide an API (in util.py or somewhere) to identify the major, minor and patch number in version
For example get_version_number(v1)  that return major, minor, path
- Provide another API which call previous API to compare version string. 
For example acos_version_compare(v1, v2), and returns >0 as v1>v2, =0 as v1==v2 or <0 as v1<v2

a10_octavia/controller/worker/tasks/vthunder_tasks.py
```diff
@@ -879,3 +891,31 @@ def _mark_lb_as_error(self, vthunder, loadbalancers_list):
...
+        compute_flag = False
+        existing_vthunder = self.vthunder_repo.get_vthunder_by_project_id(db_apis.get_session(),
+                                                                          loadbalancer.project_id)
+        if existing_vthunder:
+            compute_flag = self.vthunder_repo.get_delete_compute_flag(db_apis.get_session(),
+                                                                      vthunder.compute_id)
+        if not existing_vthunder or compute_flag:
+            try:
+                acos_version_summary = self.axapi_client.system.action.get_acos_version()
+                acos_version = acos_version_summary['version']['oper']['sw-version'][:5]
```
ytsai-a10, line 909: The version format is  [major].[minor].[patch]-[identifier]. And identifier is optional.
Can we keep them all?
[:5] is too risky, it may get wrong if major/minor number > 9

a10_octavia/db/repositories.py
```diff
@@ -350,6 +351,16 @@ def get_lb_count_by_flavor(self, session, project_ids, flavor_id):
                 or_(self.model_class.provisioning_status == consts.PENDING_DELETE,
                     self.model_class.provisioning_status == consts.ACTIVE)).count()
 
+    def get_lb_exists_flag(self, session, project_id, subnet_id):
```
ytsai-a10, line 354: can we change the API name?
get_lb_exists_flag() is confusing, because we return False when lb exists.

a10_octavia/controller/worker/flows/a10_load_balancer_flows.py
```diff
@@ -211,9 +218,17 @@ def get_delete_load_balancer_flow(self, lb, deleteCompute):
...
+    def get_new_lb_networking_subflow(self, load_balancer_id, topology):
         """Subflow to setup networking for amphora"""
-
+        lb = self._lb_repo.get(db_apis.get_session(),
+                               id=load_balancer_id)
+        vthunder = self._vthunder_repo.get_vthunder_by_project_id(db_apis.get_session(),
+                                                                  lb.project_id)
+        lb_exists_flag = False
+        if vthunder:
+            lb_exists_flag = self.loadbalancer_repo.get_lb_exists_flag(db_apis.get_session(),
```
ytsai-a10, line 229: This is just minor comment:
Since lb_exists_flag is only used for AmphoraePostVIPPlug task, can we move these check in AmphoraePostVIPPlug()?
so, we keep logic in tasks and get flow functions only add tasks in flow.

a10_octavia/controller/worker/tasks/vthunder_tasks.py
```diff
@@ -140,7 +148,11 @@ class EnableInterface(VThunderBaseTask):
     @axapi_client_decorator
     def execute(self, vthunder):
         try:
-            self.axapi_client.system.action.setInterface(1)
+            interfaces = self.axapi_client.interface.get_list()
+            for i in range(len(interfaces['interface']['ethernet-list'])):
```
ytsai-a10, line 152: Can we handle more exceptions? like dictionary key not found?

ytsai-a10 (changes requested):

a10_octavia/controller/worker/tasks/vthunder_tasks.py
NehaKembalkarA10: Done

a10_octavia/controller/worker/tasks/vthunder_tasks.py
NehaKembalkarA10: Done
Added a method `verify_acos_version(acos_version)` to incorporate this logic.

a10_octavia/controller/worker/tasks/vthunder_tasks.py
NehaKembalkarA10: Done
getting acos_version as [major].[minor].[patch]-[identifier] into db.
`acos_version = acos_version_summary['version']['oper']['sw-version'].split(',')[0]`

a10_octavia/db/repositories.py
NehaKembalkarA10: Done
Renamed it to `check_lb_with_distinct_subnet_and_project()`

a10_octavia/controller/worker/flows/a10_load_balancer_flows.py
NehaKembalkarA10: Done

a10_octavia/common/utils.py
```diff
@@ -292,3 +292,14 @@ def validate_interface_vlan_map(hardware_device):
         device_network_map.append(device_map)
     validate_vcs_device_info(device_network_map)
     return device_network_map
+
+
+def verify_acos_version(acos_version):
```
ytsai-a10, line 297: For verify_acos_version(), it can return if version is > 5.2. 
In the future if we have other feature need to do similar thing, we need to write another API to verify another specific version. So,
- can we add a constant to specify the feature version.
(for example: ACOS_DYNAMIC_INTERFACE_VERSION = "5.2.0") 
- And provide an API to compare versions.
for example: 
```
if acos_versoin_cmp(vthunder.acos_version, a10constant.ACOS_DYNAMIC_INTERFACE_VERSION) >= 0:
    ....
```

hthompson-a10: >In the future if we have other feature need to do similar thing, we need to write another API to verify another specific version.

We should not take the driver this direction. The AXAPI changes so often, between patches even, that trying to support a heterogeneous device deployment would be impossible. (I know that's not what's being proposed but it sets precedent).

We need to stick to having the acos-client deal with the versioning and enforce a homogenous device deployment. Making this driver API-aware will only lead to problems down the road beyond adding to the already high cognitive load.

For this solution, we should add a `reload_reboot` function to the `actions.py` file of the acos-client library.

```
def reload_reboot(self):
    version_summery = get_acos_version()
    version = version_summary['version']['oper']['sw-version'].split(',')[0]
    if float(version[0:3]) >= 5.2:
         reload()
    else:
         reboot()
```

a10_octavia/controller/worker/flows/vthunder_flows.py
```diff
@@ -187,6 +187,10 @@ def _get_create_amp_for_lb_subflow(self, prefix, role):
             name=sf_name + '-' + a10constants.VTHUNDER_BY_LB,
             requires=constants.LOADBALANCER,
             provides=a10constants.VTHUNDER))
+        create_amp_for_lb_subflow.add(
+            vthunder_tasks.VThunderComputeConnectivityWait(
+                name=a10constants.MASTER_CONNECTIVITY_WAIT,
```
hthompson-a10, line 192: Should this just be `a10constants.VTHUNDER_CONNECTIVITY_WAIT` we don't know the role at this point

a10_octavia/controller/worker/tasks/vthunder_tasks.py
hthompson-a10: We shouldn't be handling this KeyError. It's basically impossible to get without throwing the ACOSException error. If we are in the situation where this get's thrown, it's some catastrophic memory error that should never happen. We would want it to throw the raw exception which does still get logged.

a10_octavia/controller/worker/tasks/vthunder_tasks.py
```diff
@@ -879,3 +901,30 @@ def _mark_lb_as_error(self, vthunder, loadbalancers_list):
...
+
+class UpdateAcosVersionInVthunderEntry(VThunderBaseTask):
+
+    @axapi_client_decorator
+    def execute(self, vthunder, loadbalancer):
+        compute_flag = False
+        existing_vthunder = self.vthunder_repo.get_vthunder_by_project_id(db_apis.get_session(),
+                                                                          loadbalancer.project_id)
+        if existing_vthunder:
+            compute_flag = self.vthunder_repo.get_delete_compute_flag(db_apis.get_session(),
```
hthompson-a10, line 914: Why is this being checked? I don't think we need this right? This just tells us if there is 1 less vthunder entries. But even if there is only one, we can still just use that one's acos_version.

With this code, we end up calling this API twice per project instead of just once.

hthompson-a10 (changes requested):

a10_octavia/controller/worker/tasks/vthunder_tasks.py
NehaKembalkarA10: Removed the KeyError exception handling

a10_octavia/common/utils.py
NehaKembalkarA10: Added the `reload_reboot() `method to the `action.py` file of the `acos-client` library.
It will be risky to check like `if float(acos_version[0:3]) >= 5.2:`  ,as it may get wrong if major/minor number in acos_version > 9.
So the above specified method is modified as follows:

```
def reload_reboot(self, acos_version=None):
        if not acos_version:
            version_summary = self.get_acos_version()
            acos_version = version_summary['version']['oper']['sw-version'].split(',')[0]

        major = acos_version.split('.')[0]
        minor = acos_version.split('.')[1]
        patch = acos_version.split('.')[2]

        if major >= 5 and minor >= 2 and patch >= 0:
            self.probe_network_devices()
            self.reload()
        else:
            self.reboot()
```

a10_octavia/controller/worker/flows/vthunder_flows.py
NehaKembalkarA10: Done

a10_octavia/controller/worker/tasks/vthunder_tasks.py
NehaKembalkarA10: Removed this check.

hthompson-a10 (approved):

a10_octavia/common/a10constants.py
```diff
@@ -93,8 +93,13 @@
 CONFIRM_VRRP_STATUS = 'confirm-vrrp-status'
 ALLOCATE_VIP = 'allocate-vip'
 UPDATE_VIP_AFTER_ALLOCATION = 'update-vip-after-allocation'
+AMPHORAE_POST_VIP_PLUG = 'amphorae-post-vip-plug'
 CREATE_VTHUNDER_ENTRY = 'create-vthunder-entry'
+UPDATE_ACOS_VERSION_IN_VTHUNDER_ENTRY = 'update-acos-version-in-vthunder-entry'
 VTHUNDER_BY_LB = 'vthunder-by-loadbalancer'
+GET_VTHUNDER_BY_LB = "get-vthunder-by-lb"
+VTHUNDER_CONNECTIVITY_WAIT = "vThunder-connectivity-wait"
```
hthompson-a10, line 101: A nitpick, but the "T" here needs to be lowercase too keep consistency with the other constants

hthompson-a10 (changes requested):

a10_octavia/common/a10constants.py
NehaKembalkarA10: Done

ytsai-a10 (approved):

PR #341, opened 2021-03-15
[BUG][STACK-2166][LoadBalancer]: Unable delete LB when it is in error state.

rahulkatre-a10:
## Description
- Required: High
- Required: When VRID is not created, tried creating LB which fails causes LB to enter ERROR state. When we try to delete LB  an error is caused. The reason is vthunder object is not created before accessing it.

## Jira Ticket
https://a10networks.atlassian.net/browse/STACK-2166

## Technical Approach
- Checked the logs for deletion of LB.

## Manual Testing
- Deletion of LB which is in error state.


ytsai-a10 (approved):

ytsai-a10 (changes requested):
I think the target  branch is master not experimental/victoria

PR #398, opened 2021-05-12
[BUG][STACK-2318][vthunder: active-standby > security group] failed to delete lb --cascade in demo project

rahulkatre-a10:
- Severity Level: HIGH
- Issue Description: Unable to delete the load balancer created in the DEMO project. The issue also exists for CASCADE delete.
[vthunder: active-standby] failed to create lb in demo project and get AxapiJsonFormatError: 1 vMaster not found in vcs-summary, new error WrappedFailure. New error 1 vBlade not found in vcs-summary 

## Jira Ticket
https://a10networks.atlassian.net/browse/STACK-2318

## Technical Approach
- The list ports were getting fetched according to the project, which caused demo project ports to be left out.
-  Removed the filtering criteria for list port so that all ports get fetched irrespective of project.

## Manual Testing
- Created LB in DEMO project, and checked the delete operation for the created LB.
image

- Created LB, listener, pool, and member and checked the cascade delete operation.
image

- Created LB in admin project and checked the delete operation on created LB.
image

- Created LB, listener, pool and member in admin project and checked the cascade delete operation.
image

### DIFF
--- a/a10_octavia/network/drivers/neutron/a10_octavia_neutron.py
+++ b/a10_octavia/network/drivers/neutron/a10_octavia_neutron.py
@@ -173,13 +173,9 @@ class A10OctaviaNeutronDriver(allowed_address_pairs.AllowedAddressPairsDriver):
 
     def _get_ports_by_security_group(self, sec_grp_id):
 
-        # Handle backwards compatibility issue with neutronclient
-        try:
-            all_ports = self.neutron_client.list_ports(project_id=self.project_id)
-        except neutron_client_exceptions.BadRequest:
-            all_ports = self.neutron_client.list_ports(project=self.project_id)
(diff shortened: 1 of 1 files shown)


hthompson-a10:
Don't merge this. This method of getting ports via project was taken from code upstream and likely has an edge case that is not being addressed. Either it's that ports should only live in the admin project or that the driver's project_id isn't getting swapped properly.

Please see: https://github.com/openstack/octavia/blob/16e83eef0154a30ef91fb09842bf009c540f1c66/octavia/network/drivers/neutron/base.py#L131

hthompson-a10 (approved):

ytsai-a10 (approved):

PR #505, opened 2021-09-01
[STACK-2615]: Added forward proxy server configuration task.

rahulkatre-a10:
Note: Will add more tests for forward proxy server configuration.


hthompson-a10 (changes requested):

a10_octavia/controller/worker/tasks/glm_tasks.py
```diff
@@ -188,3 +188,32 @@ def execute(self, vthunder):
...
+            return
+        host = CONF.glm_license.proxy_host
+        port = CONF.glm_license.proxy_port
+        username = CONF.glm_license.proxy_username
+        enable_password = CONF.glm_license.proxy_password
+        if host and port:
+            try:
+                if enable_password:
+                    secret_string = CONF.glm_license.proxy_secret_string \
+                        if CONF.glm_license.proxy_secret_string else ""
```
hthompson-a10, line 208: Pretty sure this entire block can be collapsed into a single create call wrapped in a try except. We expect that the acos-client will raise an error in the event that the secret string isn't specified. We don't want to continue with creation if the secret string has been specified.

a10_octavia/controller/worker/tasks/glm_tasks.py
```diff
@@ -188,3 +188,32 @@ def execute(self, vthunder):
...
+                    self.axapi_client.glm_proxy_server.create(host, port, username,
+                                                              enable_password, secret_string)
+                else:
+                    self.axapi_client.glm_proxy_server.create(host, port, username)
+            except acos_errors.ACOSException as e:
+                LOG.error("Could not configure forward proxy server for A10 "
+                          "Thunder device", vthunder.ip_address)
+                raise e
+        else:
+            LOG.warning("Please provide forward proxy server IP address and port. ")
```
hthompson-a10, line 218: Forward proxy isn't a required setting. We shouldn't log this warning unless the other configuration information is set

hthompson-a10 (changes requested):

a10_octavia/controller/worker/tasks/glm_tasks.py
rahulkatre-a10: I was with the impression that user can use empty password. But checked on vthunder empty password is not supported.

a10_octavia/common/config_options.py
```diff
@@ -74,9 +74,12 @@
     cfg.StrOpt('proxy_username',
                default=None,
                help=_('Username used to authenticate to forward proxy')),
-    cfg.StrOpt('proxy_password',
+    cfg.BoolOpt('proxy_password',
```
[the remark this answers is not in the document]
rahulkatre-a10: Added proxy-server configuration to flavor schema

a10_octavia/controller/worker/tasks/glm_tasks.py
rahulkatre-a10: Added changes

a10_octavia/tests/unit/controller/worker/tasks/test_glm_tasks.py
```diff
@@ -433,3 +438,15 @@ def test_RevokeFlexpoolLicense_execute_no_vthunder_warn(self):
         with self.assertLogs(task_path, level='WARN') as cm:
             revoke_task.execute(None)
             self.assertEqual(expected_log, cm.output)
+
+    def test_ConfigureForwardProxyServer_execute_success(self):
```
hthompson-a10, line 442: There needs to be more tests here. Ideally each if-statement and try-except has a test written for it. It succeeding isn't really an edge case

hthompson-a10 (changes requested):

a10_octavia/controller/worker/tasks/glm_tasks.py
```diff
@@ -188,3 +188,39 @@ def execute(self, vthunder):
...
+
+
+class ConfigureForwardProxyServer(task.Task):
+    def __init__(self, *args, **kwargs):
+        self.host = self.port = self.username \
+            = self.enable_password = self.secret_string = None
+        super(ConfigureForwardProxyServer, self).__init__(*args, **kwargs)
+
+    @axapi_client_decorator
+    def execute(self, vthunder, flavor=None):
```
hthompson-a10, line 200: This needs a revert as well

hthompson-a10 (changes requested):

a10_octavia/tests/unit/controller/worker/tasks/test_glm_tasks.py
rahulkatre-a10: Added more tests

a10_octavia/controller/worker/tasks/glm_tasks.py
rahulkatre-a10: Added revert

a10_octavia/api/drivers/flavor_schema.py
```diff
@@ -208,6 +208,21 @@
...
             }
+        },
+        "proxy-server": {
+            "type": "object",
+            "description": "Forward proxy-server configuration details.",
+            "properties": {
+                "proxy_host": {
+                    "type": "string"
+                },
+                "proxy_port": {
```
hthompson-a10, line 219: These are json definitions so they need to have a dash instead of underscore. They also need a description

a10_octavia/controller/worker/tasks/glm_tasks.py
```diff
@@ -188,3 +188,46 @@ def execute(self, vthunder):
...
         self.axapi_client.delete.glm_license.post()
+
+
+class ConfigureForwardProxyServer(task.Task):
+    def __init__(self, *args, **kwargs):
+        self.host = self.port = self.username \
+            = self.enable_password = self.secret_string = None
+        super(ConfigureForwardProxyServer, self).__init__(*args, **kwargs)
+
+    def fetch_configurations(self, flavor):
```
hthompson-a10, line 199: We should keep the precedence approach to flavor vs config instead of the all-or-none approach defined here. The issue is that the user may have a single proxy host where they have multiple proxies configured on different ports. Or perhaps they want some devices to login as one user and others to login as other users.


```
proxy_fl = flavor.get('proxy-server') if flavor and flavor.get('proxy-server') else {}
self.host = proxy_fl.get('proxy_host') if proxy_fl.get('proxy_host') else CONF.glm_license.proxy_host
......
```

This gives the user more freedom

a10_octavia/api/drivers/flavor_schema.py
hthompson-a10: I had previous added these for STACK-2888 but removing them to prevent clashing. Should just be able to copy and paste this in

```
                "proxy-host": {
                    "type": "string",
                    "description": "Hostname of proxy server used for requests to GLM"
                },
                "proxy-port": {
                    "type": "string",
                    "description": "Port number used to establish connection with proxy server"
                },
                "proxy-username": {
                    "type": "string",
                    "description": "Username for proxy authentication"
                },
                "proxy-password": {
                    "type": "bool",
                    "description": "Password for proxy authentication"
                },
                "proxy-password": {
                    "type": "string",
                    "description": "Password for proxy authentication"
                },
```

hthompson-a10 (changes requested):

a10_octavia/controller/worker/tasks/glm_tasks.py
hthompson-a10: Actually found an even better way to do this since facing same issue on GLM activation.

```
def build_configuration(self, flavor):
        glm_config = dict(CONF.glm_license)
        glm_config.update(flavor.get('proxy-server'))
        
       config_payload = {
           'proxy-server': glm_config.get('proxy-server'),
           .....
       }
      return config_payload

def execute(...):
    ....
    config_payload = build_configuration(flavor)
    ....
    api_call(arg1, arg2, **config_payload)
```

a10_octavia/api/drivers/flavor_schema.py
rahulkatre-a10: Used above provided flavor schema content.

a10_octavia/controller/worker/tasks/glm_tasks.py
rahulkatre-a10: done

PR #509, opened 2021-09-09
[STACK-2888] Add further GLM flavor and config options

hthompson-a10:
## Description
Added extra glm options that were not initially added to the flavors or config file

## Jira Ticket
[STACK-2888] (https://a10networks.atlassian.net/browse/STACK-2888)

## Limitation / Known Issue
Setting "burst" as a configuration option when it is not allowed on the license causes the flow to become permanently stuck. A fix for this will be added in a subsequent PR.

## Technical Approach
- Added helper function to make comparing flavor and conf options easier

## Config Changes
N/A

## Test Cases
N/A

#### Test Config
```
[glm_license]
primary_dns = 8.8.8.8
secondary_dns = 8.8.4.4
amp_license_network = 9826ed34-b082-4db0-bbb6-e556d7a6e341
allocate_bandwidth = 25
burst = True
interval = 5
enable_requests=True
```

#### Resultant show run
image

[STACK-2888]: https://a10networks.atlassian.net/browse/STACK-2888


a10_octavia/controller/worker/tasks/glm_tasks.py
```diff
@@ -124,16 +136,31 @@ def revert(self, vthunder, flavor=None, *args, **kwargs):
 
 class ActivateFlexpoolLicense(task.Task):
 
+    def build_configuration(self, flavor={}):
```
rahulkatre-a10, line 139: we could use static method instead of instance method.

rahulkatre-a10 (changes requested):

a10_octavia/controller/worker/tasks/glm_tasks.py
```diff
@@ -161,11 +188,11 @@ def execute(self, vthunder, amphora):
                 self.axapi_client.system.action.setInterface(ifnum)
 
         try:
+            import rpdb; rpdb.set_trace()
```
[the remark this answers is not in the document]
hthompson-a10: Yup kept developing on the branch after running into the burst issue. Added "DO NOT MERGE" tag

a10_octavia/controller/worker/tasks/glm_tasks.py
hthompson-a10: We could but what's the benefit over a class method? A static method is generally used as a utility when you call the class. This is being used internally. If anything, we should turn this into a protected method since it's internal to the class.

rahulkatre-a10: This method does not use the object instance i.e, self or class  for its operation. I think there is no use accepting self parameter for this utility function.  If turned to static we can invoke it through class and object as well.

hthompson-a10: Right by why would you do that? We don't ever use it via the class

PR #517, opened 2021-09-17
[STACK-2611][STACK-2612] Initial GLM integration with flexpool support

hthompson-a10:
## Description
[STACK-2611] - Adds the ability to configure primary & secondary DNS servers of vThunder-Amphora
[STACK-2612] - Adds automated licensing of vThunder-Amphora via flexpool tokens

Currently, the vThunder-as-Amphora usage mode is not production ready in large part due to the fact that the vThunder instances are not automatically licensed. The code in this PR is capable of licensing the devices though it is largely WIP.

*Adds license revocation task which occurs just before Amphora deletion and unit test harness for glm tasks

### Bug
[STACK-2757](https://a10networks.atlassian.net/browse/STACK-2757)

Severity: High

The `_create_active_standby_amp` subflow called `GetVThunderByLB` and `vThunderComputeConnectivityWait` twice for the vMaster instead of once for the vMaster and the vBlade. This likely resulted in the errors observed in [STACK board](https://a10networks.atlassian.net/browse/STACK-2295)/PR #376. Those changes have not been reverted though as this has not been fully confirmed.

## Linked PR
https://github.com/a10networks/acos-client/pull/357/files

## Limitations
This only works with Capacity Flexpool and Capacity Flexpool burst subscriptions. Other licensing methods such as "Pay-as-You-Go" will not be supported.

## Jira Ticket
- [STACK-2611](https://a10networks.atlassian.net/browse/STACK-2611)
- [STACK-2612](https://a10networks.atlassian.net/browse/STACK-2612)
- [STACK-2615](https://a10networks.atlassian.net/browse/STACK-2615)
- [STACK-2757](https://a10networks.atlassian.net/browse/STACK-2757)
- [STACK-2759](https://a10networks.atlassian.net/browse/STACK-2759)
- [STACK-2768](https://a10networks.atlassian.net/browse/STACK-2768)
- [STACK-2790](https://a10networks.atlassian.net/browse/STACK-2790)
- [STACK-2888](https://a10networks.atlassian.net/browse/STACK-2888)

## Technical Approach
- Added task for DNS assignment
- Added task for license activation using flexpool token
- Added task to revoke license
- Added revert method to attempt to delete license if activation fails
- Corrected logic for selecting management network in both DNS and license activation tasks
- Updated and added more detail to log messages
- Fixed various styling issues
- Added helper function to make comparing flavor and conf options easier

## Config Changes
- Added `[glm_license]` as a section header
- Added `flexpool_token`
- Added `primary_dns`
- Added `secondary_dns`
- Added `allocate_bandwidth`
- Added `proxy_host`
- Added `proxy_username`
- Added `proxy_password`
- Added `proxy_port`

- Changed help text across various options to read "vThunder-Amphora" instead of "vThunder".  This is to make a distinction between the common deployment of vThunder instances on Openstack and Amphora created via Octavia.
- Added `amp_mgmt_net` to better identify the management network in the configuration file and ensure it is slotted into the mgmt interface on the ACOS device. If it is not specified then the first network defined in `amp_boot_network_list` is used.
- Added `amp_license_network` which is used to specify the network to connect to the GLM from. If this is not used, then the `amp_mgmt_net` or first network in `amp_boot_network_list ` is used.

## Flavor Changes
```
        "dns": {
            "type": "object",
            "description": "DNS nameserver information",
            "properties": {
                "primary-dns": {
                    "type": "string",
                    "description": "Primary nameserver used to contact the GLM or ELM"
                },
                "secondary-dns": {
                    "type": "string",
                    "description": "Secondary nameserver used to contact the GLM or ELM"
                }
            }
        },
        "proxy-server": {
            "type": "object",
            "description": "Forward proxy-server configuration details.",
            "properties": {
                "proxy-host": {
                    "type": "string",
                    "description": "Hostname of proxy server used for requests to GLM"
                },
                "proxy-port": {
                    "type": "integer",
                    "minimum": 1,
                    "maximum": 65535,
                    "description": "Port number through which the forward proxy server "
                                   "connects to the GLM account",
                },
                "proxy-username": {
                    "type": "string",
                    "description": "Username for proxy authentication"
                },
                "proxy-password": {
                    "type": "bool",
                    "description": "Password for proxy authentication"
                },
                "proxy-secret-string": {
                    "type": "string",
                    "description": "Password for proxy authentication"
                }
            }
        },
        "glm": {
            "type": "object",
            "description": "Global License Manager configuration settings",
            "properties": {
                "use-network-dns": {
                    "type": "bool",
                    "description": "Use the network dns nameservers instead of "
                                   "those in the config or flavor."
                },
                "allocate-bandwidth": {
                    "type": "integer",
                    "description": "Bandwidth allocated per amphora",
                    "minimum": 2,
                    "maximum": 204800
                },
                "burst": {
                    "type": "bool",
                    "description": "Enable bursting. Allows amphora to exceed allocated "
                                   "bandwidth limits. Ensures that packets never drop."
                },
                "interval": {
                    "type": "integer",
                    "description": "Interval for license manager heartbeat in hours",
                    "minimum": 1,
                    "maximum": 8760
                },
                "port": {
                    "type": "integer",
                    "description": "Port with which to send HTTP/S license request",
                    "minimum": 1,
                    "maximum": 65535
                },
                "enable-requests": {
                    "type": "bool",
                    "description": "Enables license retrieval from the GLM/ELM server. "
                                   "Allows license changes to be replicated automatically.",
                },
            }
```

### DNSConfiguration tests
- Assert warning occurs if no vthunder found
- Assert warning occurs if no networks are specified in config file
- Assert that dns is not set in the AXAPI when no dns servers are specified
- Assert that amp_license_net can be used exclusively
- Assert that setting both primary and secondary dns results in a call to the AXAPI with both nameservers
- Assert that dns nameservers can be retrieved from the subnet details and applied to the vThunder-Amphora
- Assert that a warning occurs whenever more than two name servers are found in the subnet and that the warning provides the nameservers that will be used
- Assert that amp_mgmt_net can be used exclusively
- Assert that the first boot network is chosen if amp_mgmt_net and amp_license_net are not listed
- Assert that the config file takes precedence over the network name servers
- Assert that the flavor takes precedence over the config file (therefore over network nameservers too)
- Assert that only setting a secondary name server results in a PrimaryDNSMissing exception
- Asser that a successful revert results in the dns being deleted
- Assert that no dns delete is not called during a revert when no dns servers are configured

### ActivateFlexpoolLicense tests
- Assert that warning occurs if no vthunder found
- (sanity) Assert that create called successfully
- Assert that `use_mgmt_port` is set when license net matches mgmt net
- Assert that disabled interfaces are set to up before glm license request
- Assert that hostname is set with the proper number of characters
- Assert that reverting results in the license being deactivated

### RevokedFlexpoolLicense
- (sanity) Assert that license delete is called during revoke task
- Assert that warning is logged when no vthunder is provided

### ComputeCreate
- Assert that no networks are added if none are defined
- Assert that only mgmt network is added when it is the only one defined
- Assert that only the boot networks are added if they are the only ones defined
- Assert that only the license network is added when it is the only one defined
- Assert that only the loadbalancer network is added when it is the only one defined
- Assert that only one network is added when mgmt is license net
- Assert that no extra net is added and mgmt is in pos 0 when it is the first in the boot list
- Assert that only one net is added when mgmt is lb vip net
- Assert that no extra net is added when glm is already in boot list
- Assert that no duplicates exist when net list equals sum of other net configs
- Assert that networks are added to net id list when net list is not equal to sum of other net configs

## Manual Testing
Please see tests attached to stories in the comments.

[STACK-2611]: https://a10networks.atlassian.net/browse/STACK-2611
[STACK-2612]: https://a10networks.atlassian.net/browse/STACK-2612

#### Test Config
```
[glm_license]
primary_dns = 8.8.8.8
secondary_dns = 8.8.4.4
amp_license_network = 9826ed34-b082-4db0-bbb6-e556d7a6e341
flexpool_token = <omitted>
allocate_bandwidth = 25
burst = True
interval = 5
enable_requests=True
```

#### Resultant show run
image


a10_octavia/controller/worker/tasks/glm_tasks.py
```diff
@@ -0,0 +1,253 @@
...
+        if not vthunder:
+            LOG.warning("No vthunder therefore dns cannot be assigned.")
+            return None
+
+        flavor = flavor if flavor else {}
+        primary_dns, secondary_dns = self._get_dns_nameservers(vthunder, flavor)
+        if not primary_dns and secondary_dns:
+            LOG.error("A secondary DNS with IP %s was specified without a primary DNS",
+                      primary_dns)
+            raise a10_ex.PrimaryDNSMissing(secondary_dns)
```
ytsai-a10, line 113: Can we raise only when flexpool_token is configured (for backward compatibility)?
Otherwise after upgrade the a10-octavia (or git pull) and without add dns configuration, the a10-octavia will always raise here.

a10_octavia/controller/worker/tasks/a10_compute_tasks.py
```diff
@@ -39,12 +39,16 @@ def execute(self, amphora_id, loadbalancer=None,
                 server_group_id=None, ports=None, network_list=None):
 
         ports = ports or []
-        network_ids = CONF.a10_controller_worker.amp_boot_network_list[:]
+        mgmt_net = CONF.a10_controller_worker.amp_mgmt_network
+        network_ids = [mgmt_net] if mgmt_net else []
+        network_ids.extend(CONF.a10_controller_worker.amp_boot_network_list[:])
```
ytsai-a10, line 44: Do we need to consider the case that:
- amp_mgmt_network is already in amp_boot_network_list or,
- amp_license_network is in network_ids (but not the first one)

a10_octavia/controller/worker/tasks/glm_tasks.py
```diff
@@ -0,0 +1,253 @@
...
+                if len(CONF.a10_controller_worker.amp_boot_network_list) >= 1:
+                    license_net_id = CONF.a10_controller_worker.amp_boot_network_list[0]
+                else:
+                    LOG.warning("No networks were configured therefore "
+                                "nameservers cannot be set.")
+                    return None
+
+        # ID provided to amp_license_network may be equal other provided mgmt network IDs
+        if (license_net_id == amp_mgmt_net or
+                (amp_boot_nets and license_net_id == amp_boot_nets[0])):
```
ytsai-a10, line 180: if amp_mgmt_net is configured, amp_boot_nets[0] will not be the mgmt_port.

a10_octavia/controller/worker/tasks/glm_tasks.py
```diff
@@ -0,0 +1,253 @@
...
+    def execute(self, vthunder, amphora, flavor=None):
+        if not vthunder:
+            LOG.warning("No vthunder therefore licensing cannot occur.")
+            return None
+
+        flavor = flavor if flavor else {}
+        amp_mgmt_net = CONF.a10_controller_worker.amp_mgmt_network
+        amp_boot_nets = CONF.a10_controller_worker.amp_boot_network_list
+        config_payload = self._build_configuration(flavor)
+        hostname = "amphora-" + amphora.id
```
ytsai-a10, line 163: Can we return if token not in config_payload?
Otherwise a10-octavia will always failed if no glm configured.

ytsai-a10 (changes requested):
The patch looks good to me. Just have one concern: 
- If the glm is not configured, seems the vthunder will not able to create. I think exist user/developer may have issue in vthunder flow after upgrade/update a10-octavia.

mohdadeebkhan (approved):

NehaKembalkarA10 (approved):

a10_octavia/controller/worker/tasks/glm_tasks.py
hthompson-a10: This only raises if the secondary dns is set without a primary. Even if this occurs without a token it is still not permitted.

```
        elif not primary_dns and not secondary_dns:
            return None
```
The elif handles the case where no config is set.

ytsai-a10: ok

a10_octavia/controller/worker/tasks/a10_compute_tasks.py
```diff
@@ -39,16 +39,17 @@ def execute(self, amphora_id, loadbalancer=None,
...
 
         ports = ports or []
-        network_ids = CONF.a10_controller_worker.amp_boot_network_list[:]
-        if network_list:
-            for net in network_list:
-                if net not in network_ids:
-                    network_ids.append(net)
+        mgmt_net = CONF.a10_controller_worker.amp_mgmt_network
+        network_ids = {mgmt_net} if mgmt_net else {}
+        network_ids = network_ids.union(set(CONF.a10_controller_worker.amp_boot_network_list))
```
ytsai-a10, line 44: **I think we need to make sure mgmt_net will still be the first one in the network_ids when pass to compute.build().**

Please confirm that:
When convert to set will the order be changed in the set. We need to make sure mgmt_net always be the first one in the set/list. otherwise vthunder may bootup using wrong mgmt network

ytsai-a10 (changes requested):

a10_octavia/controller/worker/tasks/a10_compute_tasks.py
hthompson-a10: Good catch. The fact that sets are an unordered datatype slipped my mind when I was writing this code. I've added some logic to fix it and a unit test harness as well for each scenario.

ytsai-a10 (approved):

mohdadeebkhan (approved):

PR #560, opened 2021-12-03
[STACK-3171] Health Monitor on target server ports

sana-a10:
## Description

- Moved health monitor from **service group** to **slb server** and attached listener port to health monitor 

## Jira Ticket
https://a10networks.atlassian.net/browse/STACK-3171
https://a10networks.atlassian.net/browse/STACK-3172
https://a10networks.atlassian.net/browse/STACK-3173
https://a10networks.atlassian.net/browse/STACK-3174
https://a10networks.atlassian.net/browse/STACK-3175
https://a10networks.atlassian.net/browse/STACK-3176
https://a10networks.atlassian.net/browse/STACK-3169

## Technical Approach

- Removed the association between **health-monitor** and **service-group**.
- Added **health_check**  parameter in axapi call of **create** and **update** slb server operations to attach healthmonitor  to slb server.
- Modified **delete_pool_flow** to execute **self._get_delete_member_vthunder_subflow** before **self._get_delete_health_monitor_vthunder_subflow** as now health monitor is attached to slb server so we cannot delete it before deleting the server associated with it for automated flow and rack flow.
- Modified **delete_pool_flow** to execute **self._get_delete_member_vthunder_subflow** before **self._get_delete_health_monitor_vthunder_subflow** as now health monitor is attached to slb server so we cannot delete it before deleting the server associated with it for loadbalancer cascade delete operation.

## Manual Testing

1. Create a load balancer

stack@automation-1:~#openstack loadbalancer create --vip-subnet-id 172.16.0.0 --name lb1

+---------------------+--------------------------------------+
| Field               | Value                                |
+---------------------+--------------------------------------+
| admin_state_up      | True                                 |
| created_at          | 2021-12-09T04:23:51                  |
| description         |                                      |
| flavor_id           | None                                 |
| id                  | 9a20812f-bc9f-46ad-8e1b-0f90de265563 |
| listeners           |                                      |
| name                | lb1                                  |
| operating_status    | OFFLINE                              |
| pools               |                                      |
| project_id          | 97181a05bddd451bb00b5da4a51686ee     |
| provider            | a10                                  |
| provisioning_status | PENDING_CREATE                       |
| updated_at          | None                                 |
| vip_address         | 172.16.0.5                           |
| vip_network_id      | ff02c2ed-de14-44d3-81d4-0f4314b57fd4 |
| vip_port_id         | 81f104e8-e11b-40dc-9284-65a67416654e |
| vip_qos_policy_id   | None                                 |
[...]

PR #571, opened 2022-05-05
Add use_override_port option

spencerharmon:
use_override_port option, as the default behavior, will always set the override port to the port of the listener associated with a health monitor, if any. Setting this to false ensures that the health monitor will monitor the members on their configured ports unless an override port is configured via flavorprofile.


ytsai-a10:
The issue we already fixed in v1.3.3 but the patch is not in victoria yet.
Here are the PRs for it: 
- https://github.com/a10networks/a10-octavia/pull/560
- https://github.com/a10networks/acos-client/pull/366
We are porting this to Victoria branch.

PR #574, opened 2022-06-28
[STACK-3250] ] Fully Populated Loadbalancer Create

ytsai-a10:
## Description
    Per the Octavia API documentation, it supports the ability to create a fully populated loadbalancer, a single request with the VIP, listener, pool, members, and health checks all in one single request sent to the Octavia api.

    The current Openstack Cloud Controller Manager code (used for Kubernetes support) as of 1.21 uses fully populated load balancers for initial service creations by default. However, 1.20 did not support using providers other than “Octavia” or “amphora” so that code doesn’t work for us, our users are currently on 1.19.

     Our users can’t use the latest code or send POST requests with fully populated load balancers despite the Octavia documentation telling them its supported, if they do there isn’t any sort of error or response to the user/API.  Instead, the load balancer and all its child objects are stuck in an immutable PENDING_ state that requires manual intervention on the hardware a10, database cleanup by hand and then Openstack port deletion to clean up.

## Jira Ticket
https://a10networks.atlassian.net/browse/STACK-3250

## Technical Approach
The detail design and descriptions are in design document: 
https://a10networks.sharepoint.com/:w:/s/Openstack/EemG6Tl7hfNGh8handWyMZwBhKjYc0ra25xqaHuqQhTMzg?e=crP4Al

## Config Changes

<pre>
<b>N/A</b>
</pre>

## Test Cases
Please reference unit-test document for detail testing steps, configuration, commands and logs:
https://a10networks.sharepoint.com/:w:/s/Openstack/ETC9KkK0VThBsY9E-mclFwgBo7FJPPGq7Ym5n7rWTJlIyw?e=I3o8n2

## Manual Testing

Please reference unit-test document for detail testing steps, configuration, commands and logs:
https://a10networks.sharepoint.com/:w:/s/Openstack/ETC9KkK0VThBsY9E-mclFwgBo7FJPPGq7Ym5n7rWTJlIyw?e=I3o8n2


a10_octavia/controller/worker/flows/a10_member_flows.py
```diff
@@ -140,6 +140,111 @@ def get_create_member_flow(self, topology):
...
+            create_member_flow.add(a10_network_tasks.ValidateSubnet(
+                name=sf_name + '_validate_subnet',
+                requires=constants.MEMBER,
+                inject={constants.MEMBER: member}))
+        create_member_flow.add(a10_database_tasks.GetVThunderByLoadBalancer(
+            name=sf_name + '_get_vthunder_by_LB',
+            requires=constants.LOADBALANCER,
+            provides=a10constants.VTHUNDER))
+
+        """
```
sana-a10, line 163: Don't we need this ?

As per my understanding we need this to attach interfaces to vThunder 

So, The code might get fail if the member is other than loadbalacer subnet or we do create second fully populated loadbalncer in project with some different network

a10_octavia/controller/worker/flows/a10_listener_flows.py
```diff
@@ -70,20 +74,74 @@ def get_create_listener_flow(self, topology):
...
-            name='check_listener_type_' + listener_name,
-            requires=constants.LISTENER,
-            rebind={constants.LISTENER: listener_name})
+
+        if listener is not None:
+            check_ssl = cert_tasks.CheckListenerType(
+                name='check_listener_type_' + listener.id,
+                requires=constants.LISTENER,
+                inject={constants.LISTENER: listener})
+        else:
```
sana-a10, line 142: Do we need this if-else check 

As per my understanding we will come to this flow if listener is present in listeners.

https://github.com/a10networks/a10-octavia/pull/574/files#diff-069d94c635a94902be690b4ff65292d39b780e0ccc6e2ce86afb425c882a7226R131

sana-a10 (changes requested):
Just a minor comment, Can we put the **_error_on_revert**  as constant value in a10constants.py and use its name to append to **sf_name** parameter.

Also can we do similar change for all other tasks and flows those are using string value directly  instead of taking it from a10Constants.py ?

https://github.com/a10networks/a10-octavia/pull/574/files#diff-753f3a82e29b521c9dcc8d6b1bbf6373ee1b0d9fe777eabbd8a45cb84b02b6d1R84

a10_octavia/controller/worker/flows/a10_member_flows.py
ytsai-a10: - For first loadbalancer creation, we need to build compute. And we add all member subnet at that time: 
https://github.com/a10networks/a10-octavia/blob/feature/fully_polulated_lb/a10_octavia/controller/worker/tasks/a10_compute_tasks.py#L63

- For instance already exist case, since the member already in database. The network delta is also handled with loadbalancer creation: https://github.com/a10networks/a10-octavia/blob/feature/fully_polulated_lb/a10_octavia/controller/worker/flows/a10_load_balancer_flows.py#L406

That means, we add member subnets when instance create or loadbalancer create. otherwise we will need to keep reloading thunder in one fully populated request.

a10_octavia/controller/worker/flows/a10_listener_flows.py
ytsai-a10: yes, handle_ssl_cert_flow() still called by other follows (other than fully populated flow).
For example: get_create_listener_flow() it will not provide the listener when calling handle_ssl_cert_flow() and the behavior should be the same as before.

PR #581, opened 2022-08-29
add notifications for zed release

spencerharmon:
https://review.opendev.org/c/openstack/octavia/+/831051


ytsai-a10:
Since not every user has the patch notification patch: https://review.opendev.org/c/openstack/octavia/+/831051
It is better to have configuration option to switch enable notification or not. And prevent using it when it is disabled.

spencerharmon:
Thank you for the review! I added a configuration option that mirrors the one we're using upstream in zed. How does that look to you?

ytsai-a10 (approved):

